PR #4, opened 2021-08-09
pls add peepclapping emoji to server 🥰

Armada99:
all original messages deleted from link queue and posted in respective channels


Armada99:
also pulled in coder's request into mine

christiankek:
shouldn’t this be something that gets suggested in the server rather then a pull request?

PR #35, opened 2021-09-14
Update GamingPiracyGuide.md

illusionX3:
don't know why we're using the more outdated emulation fandom wiki for this specifically, but emugen for everything else


Armada99:
i second

nbats (approved):

PR #40, opened 2021-09-23
Remove Vivaldi

Baba280251:
Vivaldi is a closed source unsafe browser, Even if you don't care about "Open Source", It's more important to have browsers that are safe such as Tor Browser in the beginning, Vivaldi offers nothing that makes it secure and "Private", The company actively tries to shut down any and all reverse engineering of the bad chromium skin and the code is shut down behind doors, Tor Browser on the other hand, Its source is free for public consumption, anyone can edit it as they please, and bugs can be fixed faster (if found), and just as an added bonus the browser actually masks your IP.

### DIFF
--- a/STORAGE.md
+++ b/STORAGE.md
@@ -1307,7 +1307,7 @@ Add the following commands to a search to manually scrape each site.
 
 **Multi-Platform**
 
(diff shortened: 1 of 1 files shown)


nbats:
We already have tor on the main privacy page. We're not claiming vivaldi is as secure as tor, but it has things like cookie crumbler and startpage as their search engine which is enough to call it privacy based imo so I see no good reason to remove it. It's not as secure as tor, but that doesn't make it unsafe in anyway, but for what it's worth I'll move Tor to the top of the browser tools section in the privacy wiki and also bump vivaldi to the bottom of the list in storage.

nbats:
Yeah you're right, it's not privacy focused really. We'll remove it, but we already have Tor in the main adblock/privacy section.

taskylizard:
> closed source 

Closed source =/= bad, i don't use vivaldi personally but it being closed source doesn't make it bad.

ikkqqstudio:
how about this
https://vivaldi.com/source/

ikkqqstudio:
Vivaldi source code is still avaliable, tho

PR #44, opened 2021-11-20
Update GamingPiracyGuide.md

TheGreatChanger:
On r/PiratedGames Mega Thread:
"cracked-games.org - Malware Risk."

Source:
https://rentry.org/pgames-mega-thread


nbats (changes requested):
I see, thanks for the heads up.

PR #57, opened 2021-12-23
Removal of Ungoogled-Chromium and addition of Brave

coolio321:
Removed Ungoogled-Chromium and replaced it with Brave.
Ungoogled chromium has severe security issues that Edge/Brave/Chrome don't have listed below.
https://qua3k.github.io/ungoogled/
Security is necessary for privacy.

### DIFF
--- a/STORAGE.md
+++ b/STORAGE.md
@@ -1516,7 +1516,7 @@ Add the following commands to a search to manually scrape each site.
 
 **Multi-Platform**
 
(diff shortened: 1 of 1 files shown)


nbats:
I'll move ungoogled chromuim to normal browser instead, and add Hexavalent, but I can't add Brave since people consider it to be very bloated and have apparently had some shady dealings in the past

PR #61, opened 2022-01-17
Update VideoPiracyGuide.md

illusionX3:
i think its worth pointing out as not everyone can use 1080p


nbats:
Yeah fair enough, I'll add it to the post when I update it next.

PR #62, opened 2022-01-30
Update

tslatothemoon:
steamunlocked is flagged as untrusted in piratedgames megatherthread , cause it steals rips from IGGGAMES who are full of malware


nbats:
nah IGG games is disliked for a lot of reasons, but they've never been caught actually uploading malware.

PR #78, opened 2022-03-08
Update Non-English.md

zeronegithub:


nbats (commented):
thank you, I added most changes but I moved the image hosts to the storage since they'll useful for all languages

PR #99, opened 2022-03-26
Remove Steamunlocked

EngineeredJonas:
Malware risk see r/piratedgames


nbats:
No they've never been caught with malware. They're just disliked because they have slow downloads and have had malicious ads in the past, but I don't think they do anymore. If you can provide evidence of actual malware we'll remove them, but I think their library is too good to remove over ads and a slow file host especially when people should be using an adblocker to begin with.

https://i.imgur.com/RSgEUc9.png

PR #106, opened 2022-04-05
Update base64.md

zeronegithub:


nbats:
sorry but base64 is for base64 links only

zeronegithub:
> sorry but base64 is for base64 links only

how to add a url link in base64?

nbats:
paste your link here to encode it

https://www.base64decode.org/

then give it a title like the others in the section

PR #116, opened 2022-04-18
Removed Duckduckgo

Baba280251:
Duckduckgo is not a private nor free search engine, since it begun censoring Russian searches. Now, they are censoring pirate sites and youtube-dl.
https://childrenshealthdefense.org/defender/duckduckgo-censorship-destroys-brand/
https://torrentfreak.com/duckduckgo-removes-pirate-sites-and-youtube-dl-from-its-search-results-220415/


nbats (commented):
alright I'll bump it lower on the list and pin searx instead but I don't think it needs to be removed completely

PR #138, opened 2022-06-16
Update Non-English.md

zeref-dragneel:
I remove inactive sites and sites with too many ads that annoy viewers.


nbats (commented):
I added the new sites but lots of the ones you removed looks like they're still working.

PR #151, opened 2022-07-20
Added CrackPlugins2 Megathread

FMHYHelper:


nbats (commented):
Thank you. We already have a audio plugins section in storage, so I put the megathread there instead.

PR #179, opened 2022-08-06
Fixed wiki links

FmhyHelper2:
image
hopefully i didn't mess links up


nbats (commented):
thank you but I'm not sure what the difference is. All the old links to storage still work perfectly fine for me so I'm not sure its worth sending 700+ changes through for nothing.

PR #202, opened 2022-09-30
fix: add missing file extenstion

johackim:


nbats (commented):
ty but we want to keep this file in particular raw so people can easily crosscheck URLs with it.

PR #221, opened 2022-10-14
Update ReadingPiracyGuide.md

SourceNatsu:
add picoCTF and CTFlearn


nbats (commented):
Thank you, I'm gonna close because I think they fit a bit better in Developer Tools, but I'll add both there.

PR #234, opened 2022-11-01
Update STORAGE.md

Chimira-Wright:
Added https://github.com/Tungstend/HMCL-PE


nbats:
Added to android section instead since

PR #269, opened 2022-12-04
Update VideoPiracyGuide.md

Zarch7:


nbats (commented):
closed for now because the videos won't load for me but we can add later if we can get them to work

nbats (approved):

PR #271, opened 2022-12-11
single-page: Add SideStore

humanglitch191:
Sidestore is a fork of altstore that helps it run without a server.
https://github.com/SideStore/SideStore


nbats:
added thank you

PR #285, opened 2023-01-13
Update NSFWPiracy.md

Rust1667:
-moving some sites from the general site list, to the movies sites list (for better categorization)
-added 2 movie sites


nbats (approved):
thank you

PR #286, opened 2023-01-13
Update GamingPiracyGuide.md

leybbbo:
darck repacks is no longer active and their official url also got yeeted. you can check in their discord that they're no longer maintaining the project.


nbats (approved):
RIP, thanks for the heads up.

PR #292, opened 2023-01-24
Update Non-English.md

leybbbo:
Old Persian Games is a database for games developed by Iranian developers, many of which are no longer being actively published and need preservation.

It is also a website used to preserve Persian dubs of foreign games released by Iranian publishers (For example [this Persian dub of RE4](https://oldpersiangames.org/games/resident-evil-4-goldensoftware/))


nbats (approved):
thank you

PR #306, opened 2023-02-11
Update VideoPiracyGuide.md

sinister-shadow:
Added RBTV, s365 and VIPLeague to sports streaming.


nbats (approved):
Added the first two but VIPLeague streams aren't working for me

PR #307, opened 2023-02-11
Remove AppValley and TutuBox

Lunascaped:
AppValley and TutuBox are owned by the same person, Colton Adamski. This person has repeatedly been shown to be malicious and send ddos requests to other services such as Scarlet, which is the second biggest sideloading app in the community. He accomplishs this by putting malicious scripts on his websites and using his users internet to constantly request Scarlet's website in a way that will avoid cache in order to run up hundreds in hosting bills and ultimately slow down his site. The fact he weaponizes his users internet is malicious and I don't think these sites should be on this list as it can lead to the user unknowingly being part of a botnet.

Source for the JS: https://web.archive.org/web/20230109093334/http://www.tutubox.io/ (scroll to the bottom)


nbats:
Removed both and I'll add them to the Unsafe sites mega thread as well. Thanks for the heads up.

PR #312, opened 2023-02-13
Update GamingPiracyGuide.md

sinister-shadow:
Added multiple sites, rearranged sections.

### DIFF
--- a/GamingPiracyGuide.md
+++ b/GamingPiracyGuide.md
@@ -9,30 +9,49 @@
 ***
 ***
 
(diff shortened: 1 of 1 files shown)


nbats:
closed for now but I did take note of what you changed so we may implement those sections in the future

zinklog2:
I think that's a very solid way of recategorization. we should go with this.

nbats:
yeah I agree with the game download / repacks being separate, I just want to think of a better name for the "other game site" section first.

PR #317, opened 2023-02-15
Update Non-English.md

Rust1667:
Spanish sites

I added a warning for software and games since they are mixed with media content in the same site


Rust1667:
I added 8 sites, but after better checking the download process I removed 3 of the sites.

nbats (approved):

PR #327, opened 2023-02-18
Update STORAGE.md

sinister-shadow:
Added APK Mania, ModFree, EspacioAPK, DZAPK, Android OYUN, APKModHere, APKModGet, Takemod, APKKit, Softmedal All these need to be removed from to-do in case they haven't already.


nbats:
ty I added APK Mania, apkmodhere, apkmodget, takemod, apkkit.

ModFree looks like it has fake dl buttons everywhere so I didn't add that and I think softmedal is more of a untouched apk site, but it gives me bad vibes anyways. EspacioAPK + androidoyun are non english so I kept in tested.

PR #330, opened 2023-02-19
Update AdblockVPNGuide.md

Rust1667:
PowerTunnel is already in the wiki, but IMO its worth mentioning it as an alt to GoodbyeDPI (since GoodbyeDPI works only on Windows)


nbats:
True but we generally try to stick to windows only stuff for the main sections if we can.

PR #351, opened 2023-03-04
Update ReadingPiracyGuide.md

Zarch7:


nbats (commented):
ty but its already added just without the /home

PR #372, opened 2023-03-10
Added y2monkey in userscripts

jijirae:
I made a Violentmonkey script which enables the user to press "SHIFT + D" which opens the video they're watching in y2mate so they can download it.


nbats (approved):
Nice thank you

PR #374, opened 2023-03-10
Added r2monkey in userscripts

jijirae:
Same with my "y2monkey" userscript, **but for Reddit videos.**

Should I consolidate them in one line or keep it as is?


nbats (approved):
Separate is fine, thanks for adding

PR #404, opened 2023-03-20
Update base64.md

TerrenceTingleberry:
Name: Modded Spotify APK
Original: https://forum.mobilism.org/search.php?st=0&sk=t&sd=d&sr=topics&keywords=Spotify&sf=titleonly
Changed to: https://github.com/xManager-App/xManager
Reason: Automatically fetches latest APK.

### DIFF
--- a/base64.md
+++ b/base64.md
@@ -1306,7 +1306,7 @@ pass: paluch
 
 #### Modded Spotify APK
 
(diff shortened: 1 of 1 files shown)


lenicyl:
xmanager is already in the wiki, under their old repo name (https://github.com/xManager-v2/xManager-Spotify)

nbats:
yeah I'll replace the starred one we have with xmanager, but I still think keeping the base64 mobilism link is worth it too.

PR #462, opened 2023-03-28
made subsections in dev-tools

zinklog2:
WIP. reordering within the subsections and changing desc is left. also  some subsections have few links but are big enough topics that they would fill up quickly now that people can see whats missing.


nbats:
Alright closed because of conflicts but I did combine everything

PR #513, opened 2023-04-14
Add new sites by GaijinRyman

GaijinRyman:
It have a whole backlog of "changed stuff" but I checked myself, no sites are missing, just added, vscode is nuts hahaha


nbats:
Sorry but there were so many edits that its conflicting with other edits. I can merge once they're fixed.

PR #525, opened 2023-04-15
Add donkuri to japanese learning resources

GaijinRyman:
Donkuri is a resource the try to gather all japanese immersion communities in one single link developing their own think, i believe it's great, since it can work to all languages too.

I also added the link to /piracy new's megathread


nbats (approved):
thank you

PR #543, opened 2023-04-20
Update VideoPiracyGuide.md

Agent-324:
corrected the zoro script name to 'auto focus on video player to use keyboard shortcuts'.


nbats (commented):
We need to keep it short so I'm gonna switch it to "Auto-Focus"

PR #554, opened 2023-04-22
Update GamingPiracyGuide.md

Kai-FMHY:


nbats:
moved to storage with other text adventure sites

PR #562, opened 2023-04-23
Update EDUPiracyGuide.md

Kai-FMHY:
Removed * [Backyard Chickens](https://www.backyardchickens.com/) - Chicken Raising Forum. I'm moving it to Gardening/Agriculture


nbats:
I don't think most people think of raising animal and gardening as the same thing really

PR #597, opened 2023-05-04
Update VideoPiracyGuide.md

karkass1:


nbats (commented):
ty but already posted in french section

PR #603, opened 2023-05-07
Update single-page

KennyPhan123:
Libgen.fun is not the main site of library genesis, I tried to search a book with this ISBN "9781426223518" I cound't find it on libgen.fun, but in libgen.is it worked.

Go to this site "https://whereislibgen.now.sh/" (in the Libgen mirrors section in fmhy web), it says that the original website for libgen is libgen.is


nbats:
thank you but don't edit the single page please, edit the individual ones.

PR #623, opened 2023-05-21
Update AI.md

defineprogramming:
Added MusicLM to AI (Audio Generators)


nbats (commented):
ty, we don't add waitlist stuff but I'll put it in our list of things to wait for.

PR #644, opened 2023-05-30
Update AI.md

zinklog2:
Merge between gpt and gpt-tools might be a bit iffy so you can see if its making sense. otherwise looks good to me

### DIFF
--- a/AI.md
+++ b/AI.md
@@ -11,69 +11,49 @@
 
 # ► AI Chatbots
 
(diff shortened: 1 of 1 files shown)


nbats (approved):
Yeah I'm fine with open source + chatbots but tbh there's not enough non-GPT based chatbots to warrant us having a GPT section. It makes more sense to just have chatbots as one, at least for now til we collect more non-GPT chatbots.

PR #645, opened 2023-05-30
Update VideoPiracyGuide.md

zinklog2:
Added resolution to ddl sites, might need to reorder the links


nbats (commented):
thank you but I think we should wait til we do all of them before sending the PR, that way some sites don't look good or bad compared to other unfairly.

Also I think it should be labeled more like this: 

Movies / TV / Anime / 1080p

PR #651, opened 2023-06-01
Update DownloadPiracyGuide.md

hussainbiedouh:
add romansrevenge.com


nbats (commented):
ty, we have to do some scans first to be sure its safe but we'll check it out.

PR #652, opened 2023-06-01
Update Non-English.md

Trem1151:


nbats (commented):
ty but the play won't work for me

PR #662, opened 2023-06-07
Multiple pages edited

enzomtpYT:
Added Hitmoi Downloader To Storage as Multi-downloader
Added Sharewood To storage as french torrenting
Added Animevost.fr To storage as french streaming
Added Photopea To Image tools as Image editing


nbats (commented):
Thank you, added the french sites and hitmoi, but photopea is already added under image editors

PR #671, opened 2023-06-12
Remove ripped.guide

Deletecat:
image
😬 

Best to remove this for now, I am not sure what sparked this but I think someone will probably know about it


nbats:
Yeah we've both removed ripped and add them to our unsafe list for doxxing. Thanks for the heads up.

Deletecat:
> doxxing

Oh wow. Didn't realise it was that bad. what a shame

PR #678, opened 2023-06-14
Update ReadingPiracyGuide.md

Kai-FMHY:
changed it from "servers" to "server" since its a singular discord server now that great e library server is removed


nbats (approved):
ty, didn't merge because I'm editing but I did add the change

PR #702, opened 2023-06-25
Create JobBoard.md

Nithur-M:
Added a job board


ggtylerr (changes requested):
you dont need a whole section let alone page just for 1 link

Nithur-M:
> you dont need a whole section let alone page just for 1 link

Hi, may I know what section I can add?
Also, I think creating this page will encourage others to add similar sites, no?

nbats:
You could add it to the career tool section

https://github.com/fmhy/FMHYedit/blob/main/TOOLSGuide.md#-career-tools

PR #705, opened 2023-06-26
Update TOOLSGuide.md

Nithur-M:
Added a description to Real Work From Anyhwere


nbats (approved):
Thank you but I'll probably shorten it a bit to something like this: 

Global "Work-From-Anywhere" Job Board

PR #710, opened 2023-06-28
Update UnsafeSites.md

Deletecat:
Replace my comments (u/kitcat976) with rentry links. I have started to go through and delete my old posts and comments, mostly because I was active in r/flashcarts and information changes all the time. Though that also means that the comments from me on the unsafe site list will disappear.

I have backed up those comments on rentry and replaced the links. The Qoob entry doesn't really need it tbh, you.could change it to also mention how they were banned from rutracker and link to that instead

I have also backed up most of the other posts on web.archive.org // archive.is // rentry, though they haven't disappeared yet so it probably won't matter too much


nbats (approved):
looks good thanks for doing that

PR #718, opened 2023-07-02
Update VideoPiracyGuide.md

angrybutter:
dbgo.fun seems down and redirects to ads, added superembed api


nbats (commented):
closed because dbgo is still up, but I did add superembed

nbats:
actually nvm dbgo is dead my bad

PR #719, opened 2023-07-06
Update UnsafeSites.md

Deletecat:
* Add note in MRepo section
* Add TotalAV / PC Protect / Protected to the list - Classed as a PUP by Malwarebytes, somewhat shady business practices, Protected owns both TotalAV and PC Protect which are both classed as PUPs (literally the same program but renamed :p )


nbats (approved):
thank you

PR #720, opened 2023-07-06
Update description for tutorialbar.com

dinesh-58:


nbats (commented):
ty, I'll move it to our udemy coupons section

PR #721, opened 2023-07-06
Debloat STORAGE.md

qiracy:
- Removed non eng search engines, will add them in non eng since they fit there better and non eng already has search engines listed for some languages
- Removed Android Browser Wiki because it was just a wikipedia page comparing all web browsers
- Removed all calculator sites except starred ones since they have all the tools one would need for any kind of calculation (including lemmy registration question)
- Removed some calendar sites due to their free plan being pretty limited, requiring sign up to even see the site, or lack of updates
- Removed all character counting tools/sites except the rentry list of them. They're pretty much the same exact thing with different UI, and the rentry list has more than enough links to them

### DIFF
--- a/STORAGE.md
+++ b/STORAGE.md
@@ -71,8 +71,6 @@
 
 **Non-English**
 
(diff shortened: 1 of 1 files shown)


nbats (commented):
ty, I removed the character counters, calendar sites, unstarred the browser list, and we can add moving the non eng search engines to #to-do. 

As far as removing the calculator sites go, I see no reason really. The sites have lots of different types of calculators / UIs.

PR #724, opened 2023-07-08
Update STORAGE.md

shahzaibshazoo:
Some of the latest updates are done to these files, old URLs are redirected to the new once that's why I added them directly


nbats (commented):
alright ty, we'll check out the temp mail site and star it if we think its good.

PR #726, opened 2023-07-09
Update GamingPiracyGuide.md

Deletecat:
Possible future edits/removals:

* LegendaryRepacks - Discord link in the wiki is dead, I haven't found a new and updated link, it's the only way to access their stuff iirc. Whether or not there is a new link, I don't know.

Removed:

* ROMsUniverse - Downloads are dead (file not found)
* LoveROMs - Fake download buttons / riddled with pop up ads for non-adblock users - sometimes clicking the page in the wrong spot will pop up an ad. The real download button(s) are in small red text which is easy to miss. Site is likely unaffiliated with the original LoveROMs which closed a few years ago

Fake download link built into LoveROMs:

image

Real download links for LoveROMs (Download, DiscX, etc):

image


nbats (approved):
alright ty for removing these, and I'll see if anyone can find a new LegendaryRepacks link

PR #730, opened 2023-07-12
feat: add yt-dlp as youtube tool

lavafroth:
Adds yt-dlp to Youtube Tools.
yt-dlp is a commandline multitool to download videos from youtube and other platforms.


lavafroth:
Duplicate, whoops! Already there on the single page.

PR #735, opened 2023-07-14
Remove nsw2u from STORAGE.md

wizpizz:
nsw2u, the Nintendo Switch ROM website has been taken, therefore I've removed it from STORAGE.md.


nbats:
seems to still work for me

PR #760, opened 2023-07-25
Update VideoPiracyGuide.md

GuiltiaSinJ:
1. Added KickAssAnime Mirrors
2. A couple of minor fixes plus link update
3. Added Anix (https://anix.to/) It's basically Animixplay but with a some UI tweaks though  barely noticable, and couple added features, but it's still the  same experience, the standout part though is that it (at least  currently) has no ads as far as I could tell


nbats (approved):
ty

GuiltiaSinJ:
no prob

PR #772, opened 2023-07-30
Update AudioPiracyGuide.md

Rearview2490:
This is the only complete list of Hardcore History episodes that I've found. This is from my personal files.
Everywhere else isn't updated fully. 

I think the content of this podcast justifies a star. 

One of the most popular podcasts on the internet. If you feel that it's not star worthy, that's okay too. As long as it's accessible to everyone. :)


nbats (commented):
Sorry we can't post direct link to pirated content.

PR #777, opened 2023-07-31
Update AI.md

suvodeep12:
Added Phind AI


nbats:
ty, it was already in dev tools, so I just moved it from there to here.

PR #784, opened 2023-08-01
Added Smutbase to NSFW

Dinopirate06:
Smutbase is a sister website to Open3DLab & SFMLab (both of which are already on the Wiki) with the main difference on this sister site being primary NSFW Models


nbats:
ty, you don't need to edit the single page so I can't merge this, but I will add the site you added.

PR #785, opened 2023-08-01
Update NSFWPiracy.md

zinklog2:
Reorganized the entire section to add more links


nbats:
Sorrry but I think its best to wait til we actually take on NSFW as a whole before we start adding empty categories. We will likely add more sections, but we just need to actually start collecting before we try to come up with them.

zinklog2:
Ahh oops i forgot to change it to draft, i meant to work on this branch for a while. Can you open it again

PR #787, opened 2023-08-02
Update GamingPiracyGuide.md

spIoopy:
added a known website with arcade dumps and games


nbats (commented):
thank you but we'll need to test it ourselves first before I can add.

spIoopy:
alright, take your time, account not required

PR #791, opened 2023-08-03
added app db

DevShubam:
I would also suggest it being starred since it's pretty much the largest app store and it has integrations with other file hosting and sideloading services


nbats (commented):
we need to scan some stuff first but we'll add soon if they come back clean

PR #793, opened 2023-08-03
Adding ARMGDDN 3D movies

Vojtak42:


nbats:
thank you, we need to test it first to be sure it works, but I'll merge this after we do.

Vojtak42:
You actually already have it in Games section.

PR #806, opened 2023-08-08
Update MISCGuide.md

Rearview2490:


nbats (commented):
thank you, I added the parking one, but I feel like the UK shop one is a bit too niche.

PR #809, opened 2023-08-08
Update GamingPiracyGuide.md

Rearview2490:


nbats:
ty but I'm going to put it next to modrinth itself in storage

PR #810, opened 2023-08-09
Update GamingPiracyGuide.md

Rearview2490:


nbats:
Sorry but when it comes to tools like this I think we need to stick to games that are more popular (like LoL, smash, cod etc.) Otherwise we'd have to do it for hundreds of games probably.

PR #814, opened 2023-08-09
Update LinuxGuide.md

Rust1667:
Categories for the Linux Tools / Apps section


nbats:
Thank you. I'll need to adjust some stuff so it's gonna be awhile before its done, but I'll refer to this while I'm working on it.

PR #822, opened 2023-08-10
Update Non-English.md

Kai-FMHY:
Removed https://www.1tamilmv.news/ it appears to be dead/down at the moment which sucks since it was starred, perhaps we should add it to to-do and watch if it will come back up.


nbats:
removed ty

PR #876, opened 2023-08-29
Update UnsafeSites.md

DogancanYr:
1337x Caught with [malware


nbats:
Yeah thank you we're aware, but 1337x did delete it pretty quickly after it was reported so idk if its fair to add them here really.

DogancanYr:
Ok.

PR #884, opened 2023-08-31
Update Android-iOSGuide.md

farooqwani:


nbats:
thank you, we'll check it out and add it if it works and is safe.

PR #885, opened 2023-09-01
Update Android-iOSGuide.md

suvodeep12:
Added AeroInsta as Instander is buggy for some people.


nbats:
thank you, we need to scan it before we can add but we'll do it soon.

PR #890, opened 2023-09-03
Added two websites to download & developer tools section

SarkarKurdish:
Developer Section:
Added transluna.net which is used to Translate json files into other languages

Download Section - File Sharing Services:
Added filedoge.com free unlimited file sharing

Both are free with no ads


nbats:
Thank you added both

PR #894, opened 2023-09-04
edit microsoft edge removal guide

azgaresncf:


nbats:
ty but we need to test it ourselves before adding

azgaresncf:
okay np 👌🏻

PR #935, opened 2023-09-17
Replace 50TB with OlaMovies

nullish-cat:
50TB no longer exists and is currently still on the beginner's guide. PR replaces it with OlaMovies (feel free to change it with something else)


nbats:
Done ty

PR #940, opened 2023-09-19
Updated predb urls

pr0pz:
- Changed predb.de to predb.net (new domain)
- Removed predb.live (not working at least for months)


nbats:
Alright thanks for the updates, but just a heads up you don't need to edit the single page, it updates automatically.

PR #947, opened 2023-09-24
Added AllMacWorlds, set AppsTorrent to star and added notice for it

rastiqdev:


nbats:
Alright thank you, I'll star AppsTorrent, but we need to test / scan files from AllMacWorlds ourselves first before adding it.

PR #956, opened 2023-09-29
Update VideoPiracyGuide.md

PickleSour:
updated this list as well


nbats:
Thank you I got it added earlier when you sent it to divolt. And also keep in mind the single page is updated automatically so you never need to touch that.

PR #958, opened 2023-10-02
Update VideoPiracyGuide.md

GuiltiaSinJ:
1. Starred Anix, I think it desrves it, the interface is good and it has a large library since it mirrors Aniwave, it's basicaly Aniwave but different design and layout. it's the same with the alternatives I added they are the same as Aniwave but with different UIs

2. Added Aniwave Lite versions which offers a more simplistic design of the original websites which are pretty good. Also updated the mirror links

3. Also unstarred Animension I think it's not all that good since it's basically a proxy player for GoGo and the biggest gripe I have is about the player being on a different domain which is a minus from me, but i like it for the feature to provide links to official streams but it's not always

4. Added more Animixplay clones with different feature like the players for example


nbats (approved):
thanks for the updates, I made a few slight changes but everything you added should be there now.

PR #960, opened 2023-10-03
Update STORAGE.md

dcolewyatt16:
Song Lyrics added


nbats:
Thank you but we need things for a section before we can make it.

PR #965, opened 2023-10-05
Update GamingPiracyGuide.md

dcolewyatt16:
Flashpoint


nbats:
Thank you, we have their old link posted already so I'll replace it with this.

PR #972, opened 2023-10-08
Update STORAGE.md

dcolewyatt16:
Added Lyrics


nbats:
thank you, we'll need to dig through them ourselves to make sure they're worthwhile, but they should be added within the next week or so.

PR #975, opened 2023-10-09
Update STORAGE.md

dcolewyatt16:
Added more lyrics


nbats:
thank you, I appreciate the amount of effort here but I gotta be honest the odds of us adding all of these is pretty low. I think its best we pick out some of the best ones and post those.

PR #979, opened 2023-10-11
Adding nunify as a video chat ptovider

mlnkrish:


Rust1667:
this seems paid, correct me if Im wrong
we only add free sites, with rare exceptions

nbats:
Yeah if its paid then we won't add it

PR #987, opened 2023-10-12
Update VideoPiracyGuide.md

oddmario:
Added a site that provides free instantly activated IPTV trial


nbats:
ty but we don't add trials for the most part

PR #989, opened 2023-10-14
Update AdblockVPNGuide.md

spIoopy:
Added a rentry page that tells you how to remove that new youtube anti adblock pop up nag with an uBlock Filter


nbats:
very nice thank you

nbats:
Alright someone said ublock already has all of these built in, so not sure there's any need to add this.

PR #1018, opened 2023-10-23
Update LinuxGuide.md

Dweep4269:
Added MacBB and MacSerialJunkie for Mac Apps and games as i think they are good websites.


nbats:
Thank you but MacBB is members-only, and macserialjunkie asked us not to include their site on our wiki sadly.

PR #1023, opened 2023-10-25
Update GamingPiracyGuide.md

dcolewyatt16:
add triviaz


nbats:
ty, added to the fun sites section with other trivia sites

PR #1026, opened 2023-10-26
Update GamingPiracyGuide.md

dcolewyatt16:
add waitwait


nbats:
ty but I feel like this is a bit too niche to add

PR #1031, opened 2023-10-26
Update single-page

dcolewyatt16:
siteindices


qiracy:
1. Never edit the single page, it gets automatically updated every 12h.

2. Try to find proper categories for the things you want to add. Look through the wiki to find a category that best matches your links rather than just dropping all your random stuff in random categories.

3. If no category seems suitable for your links, chances are they're not suitable for the wiki in the first place. Please try to only add things that you believe will be useful for other people as well, not just your personal favorite sites. It would also be nice if you could provide your reasoning for why you think something should be added to the wiki, or at the very least add a description so we know what the site is about.

taskylizard:
Closing, feel free to open a new issue with the above mentioned points in mind.

nbats:
yeah its fine I'll just add em to the link queue

PR #1050, opened 2023-10-30
Update MISCGuide.md

ballszu:
Added and starred Hacker News


nbats:
ty but we have this in the hacker news storage section

ballszu:
ok.. But this is specifically news section...

nbats:
its just under security now which ig is confusing but I'll add hacking to the name

ballszu:
all cool

PR #1059, opened 2023-10-31
Update AdblockVPNGuide.md

annemedia:


nbats:
ty but we'll need to research it a bit first before adding

annemedia:
We've launched AriaVPN recently so there's not much you can find. I'm just looking for ways to build our online presence and would appreciate it if you could add the reference. May I be of any assistance in helping with your research?

nbats:
Yeah its tough because there's so little to research, so its hard to convince people its worth adding with so little to go on. Is there any way I could get access to it for free for a day just so we can actually try it out ourselves?

nbats:
also so sort of proof that you don't log would help a lot

annemedia:
Your backlink is valuable, so I'd like to offer you more than just a one-day access. If you want, I can give you a promo code with a 99% discount on our 1-year plan. Buying a plan through our website would give you a full customer experience, including the ability to switch a server/location. You can apply the code on our website, and I'll send you ahead the differential amount in Monero or BCH to your wallet. In the alternative, I can just generate the certs and send you OpenVPN configs for all of our servers. As for logs, would screenshots of our VPN servers configs & DNS servers set up satisfy the proof? Please hit me up at support@ariavpn.net and let me know which option works best for you.

PR #1077, opened 2023-11-03
Update GamingPiracyGuide.md

dcolewyatt16:
wuht


nbats:
ty, we'll need to check them out ourselves but I'll add them to our link queue

PR #1090, opened 2023-11-04
Update STORAGE.md

aviv926:
https://github.com/NoName-exe/revanced-extended

As far as I know this is not legit and should neet to be removed from Fmhy More info here:
https://www.reddit.com/r/revancedextended/comments/17npedn/revanced_extended_app_on_github/


nbats (approved):
alright thanks for the heads up, I'll get it removed

PR #1096, opened 2023-11-06
Update GamingPiracyGuide.md

DogancanYr:
It is safe because it uses Gog installation.


nbats:
Yeah we used to have it, but I was told it was filled with adware.

DogancanYr:
> Yeah we used to have it, but I was told it was filled with adware.

Is it on the site or in the installation?
Because gog games has game ads in the installation menu.
Can I look at that evidence?

nbats:
Yeah tbh I don't have much info, thats just what I was told in the past. I'll add it to our to do list and we'll research it further.

PR #1109, opened 2023-11-11
fixed slavart main link

McD1224:
Fixed the slavart main link to discord, old one did not work, source is the divolt server


McD1224:
wrong link was used

PR #1125, opened 2023-11-13
Update VideoPiracyGuide.md

GuiltiaSinJ:
Removed animixplay.fun it's not very good and does not feel as polished as the others I don't think it deserves to be here especiially when there are as many copies of aninimix as there

Also I was wondering if it's possible to add a seperate section for streaming and downloading apps on the VideoPiracyGuide I think that would be good as apps can be more convinient and comfortable to use coz of the customisation you can't find on websites, there seems to be a lot of good apps out there but don't seem to be here so many that I think they deserve their own section, I thought I should ask first before making any major changes


nbats (approved):
yeah it depends on what you mean by apps, we do have a apps section in mobile, and also a torrent apps section in video for things like stremio, but I'm not sure if there are enough desktop streaming apps for an entire section.

GuiltiaSinJ:
Oh, okay, I see, then I guess I'll look into some more apps that are worth adding,  and if I come up with some good results I'll come back again

PR #1135, opened 2023-11-14
change homepage "browse collection" link to beginners-guide

TheScreechingBagel:
it currently links to adblockvpnguide, i think the beginners guide makes a lot more sense and is more relevant to the average new user


nbats:
I see what you're saying, my only issue is that the beginners guide isn't really part of the "collection" so to speak. Its more of its own separate thing.

PR #1145, opened 2023-11-17
Update

the9655a:
- Name corrections
- Added uBlock guide
- Removed old links
- Moved CTTs Windows Utility under System Tools. It doesnt primarily focus on debloating, and is convenient for setting up new systems, so I feel it fits better there.


nbats (approved):
thanks looks good but I did remove the ublock video as its from 2018

the9655a:
> thanks looks good but I did remove the ublock video as its from 2018

A lot of the content in the video still holds true to this day. It explains some parts of uBlock (remains the same to this day) that the other guide doesn't include. To be fair, the parts I'm referring to are more advanced, though still relevant.

nbats:
alright fair enough, I'll get it added and we'll see if anyone objects or sends something more recent

the9655a:
> thanks looks good but I did remove the ublock video as its from 2018

A lot of the content in the video still holds true to this day. It explains some parts of uBlock (remains the same to this day) that the other guide doesn't include. To be fair, the parts I'm referring to are more advanced, though still relevant.

> alright fair enough, I'll get it added and we'll see if anyone objects or sends something more recent

Cool, sounds good :)

the9655a:
> > alright fair enough, I'll get it added and we'll see if anyone objects or sends something more recent
> 
> Cool, sounds good :)

I've also personally used the guide not that long ago and I found it helpful.

PR #1149, opened 2023-11-18
Update AI.md

veryamazinglystupid:
Add sites that are useful, remove sites that aren't good.

### DIFF
--- a/AI.md
+++ b/AI.md
@@ -13,14 +13,15 @@
 
 # ► Open-Source LLMs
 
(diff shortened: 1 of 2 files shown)


qweered:
Do not remove openllm leaderboard

nbats:
Thank you, I applied most changes, but mancer looks mostly paid so I'm not sure it really fits, and I also kept the leaderboard as someone here asked us to. Also keep in mind you don't need to edit the single page.

PR #1169, opened 2023-11-26
fix: Remove duplicate mpc-hc link

lenicyl:
- The main "MPC-HC" and "Fork" link are basically the same, there is no need for "Fork" to exist


nbats:
ty, closed because I'm currently editing a new version of this page but I did update it there

PR #1175, opened 2023-11-27
Update LinuxGuide.md

Rust1667:
add a internal link to the general terminal section in the linux section (a lot of things in that section are mainly for linux)


nbats:
Thank you, but to be honest we should just move anything thats for linux out of the section into linux somewhere. I'll add doing that to our to-do list.

nbats:
Yeah I think its better to just move the linux based stuff out of system into linux instead, added to our to-do list

PR #1187, opened 2023-11-29
Update Non-English.md

qiracy:
**Changes made:**

- Added video quality and sub/dub tags to russian sites
- Updated descs and links, removed dupes and re-organized stuff where needed
- Removed https://github.com/greatfire/wiki, it's written in chinese so not much use for it in the russian section
- Removed [TLauncher Legacy](https://tlaun.ch/), no reason to keep it in russian section and it has also been dmca'd so downloads are disabled
- Removed [ШрифтыПаттерныКисти](https://t.me/+Ax8cZJCY19k3NDRi), expired link
- Removed **[Yohoho](https://yohoho.cc/), [2](https://4h0y.gitlab.io/)**, requires donation to ukraine of any amount to watch anything, and while you can donate as little as 0.1$ it's still paid and wouldn't really fit the wiki? perhaps it can be added back with a note since the donations are for a humane cause but shouldn't really be starred either way
- Removed [KinoGo](https://kinogo.la/), dead
- Removed [RussianFilmHub](https://russianfilmhub.com/), only has links to amazon and tubi


DogancanYr:
I can download from TLauncher Legacy.

PR #1195, opened 2023-12-02
Update GamingPiracyGuide.md

jorgeperezlara:
Added Miniplay and Spatial.io


nbats:
Thank you, we need to test them out ourselves first but we'll get them added after we do.

PR #1200, opened 2023-12-02
Move Mac Code Editors to the Mac Apps section

qiracy:
- Removed Mac Code Editors from storage and moved them to the Mac Apps section
- Replaced the "or" between 3+ links with an "and". Decided to keep the "or" between 2 and 3 links since it looks slightly better than "and"


nbats:
thank you I'll get the code editors moved, as far as changing or to and, or slightly shorter so I'm not sure if changing it to and is really worth it

PR #1210, opened 2023-12-06
Add 2 new Online Generators to wiki

cmartinezv:


nbats:
ty we'll need to check both out ourselves first but we'll get them added after we do.

cmartinezv:
Hi nbats ! Could you give us some details why this branch hasn't been merged so we can improve in the future ? Many thanks !

nbats:
No reason, we just have hundreds if not thousands of sites to go through so it'll take some time.

cmartinezv:
> No reason, we just have hundreds if not thousands of sites to go through so it'll take some time.

OK! Thanks, I thought it got rejected. Let me know if I can help somehow :)

nbats:
Yeah I haven't taken a very close look, but at first glance they seem like free trial style sites which we usually don't add here.

cmartinezv:
> Yeah I haven't taken a very close look, but at first glance they seem like free trial style sites which we usually don't add here.

Upps! I'm sorry then... we can't afford to give more than 3-5 images for free at this time due to the high GPU costs. I've seen projects like interiorai.com ( that doesn't even have trial I think...) and I thought I could add my project there.

Anyway if by any chance you think it could fit I'll be very gladful, if not, thanks anyway :)

PR #1211, opened 2023-12-07
Update

the9655a:
- Moved Virtual Machines section under System Tools (i think it fits better, lmk what u think)
- Added Windows Tools heading
- Updated old Tools links


nbats (approved):
Thank you looks good. I added all changes except adding the "windows tools" section as system tools is supposed to already imply windows.

PR #1225, opened 2023-12-13
Update GamingPiracyGuide.md

wongbener:


nbats:
thank you but we usually try to add browser gaming sites that are a bit more unique, pretty sure there's tons like this one.

PR #1234, opened 2023-12-15
Update storage.md

the9655a:
Improved MC Launchers section:
- Added labels
- Removed "Workarounds", no longer needed.
- Re-ordered list. Placed Prism Launcher higher, one of if not the best launchers available.


the9655a:
- Starred Patch Reddit Clients
- Mirrored with same wording to beginners guide
- Removed Infinity builder. Useless with new revanced update, less convenient, old, and breaks a lot.

PR #1245, opened 2023-12-18
Update AdblockVPNGuide.md

AbdulShan:
(webshare.io)This website Gives 10 Free Premium Proxies with 1GB/month. It is also possible to create multiple accounts using TempMail.


AbdulShan:
(https://www.webshare.io/features/free-proxy) - 10 Free Premium Proxies with 1GB/month. Can create multiple accounts using TempMail.

nbats:
thank you but finding free proxies is easy so idk if there's any need to add a site that requires signup

AbdulShan:
I had connectivity Issues with other free ones. Also seems like 1gb/month is very less.

nbats:
alright I'll add it to our list to test and see what others think

PR #1246, opened 2023-12-18
Update STORAGE.md

ballszu:
* Removed many temp mail sites all were just repetitive and some did not even work.
* Starred inboxes, works fine no problems.
:sleeping: :sleepy: :sleeping: :sleepy: :sleeping: :sleepy: :sleeping: :sleepy: :sleeping: :sleepy:


nbats:
we can star inboxes but I don't think axing this entire section is a good idea. It can think of a lot of situations where having a bunch of temp mail sites could be useful.

ballszu:
bro but they all are the same lol, literally no difference (some are worse actually)

nbats:
They're all different address', if you're trying to use a fake email to signup for a site having a bunch of options could be useful.

PR #1247, opened 2023-12-19
Update AI.md

suvodeep12:
Added LM Studio. Can consider starring?


nbats:
Possibly, but we'll need to test it ourselves first.

PR #1260, opened 2023-12-21
moved various sections from storage to where they belong better

zinklog2:


nbats (approved):
thank you very nice job

PR #1289, opened 2023-12-28
Update AudioPiracyGuide.md

Rust1667:


SnowzNZ (commented):
Pointless? You end up in the same place anyway.

PR #1294, opened 2023-12-30
Took Drive Formatting/File deletion out of Storage

MothiMediocreJoker:
Put in File-Tools.md as per nbats To-do List.


nbats:
Thank you but all of this stuff also needs to be organized into their own lines, with labels added. I should of been more clear in the to-do, sorry.

PR #1306, opened 2024-01-02
Added tags to online text editors as mentioned in to-do.

BasedF:
- Rearranged the links to look better
- Added both local and cloud saves to the links
- Removed Emacs' Org Mode as it isn't an online text editor as far as I know


nbats:
nice work, thank you for doing this

PR #1312, opened 2024-01-04
Update Android-iOSGuide.md

mtsbh:
Remove Sponsorblock link. It's NewPipe and its been archived on Dec 29.


Rust1667:
main new pipe repo isnt archived(https://github.com/TeamNewPipe/NewPipe/)
its only the "newpipe with sponsorblock" fork that was archived, and it can be replaced https://github.com/TeamNewPipe/NewPipe/, which is another fork that contains Sponsorblock too

PR #1318, opened 2024-01-04
Update Non-English.md

iamgpe:
Added hotstream.me to french streaming websites


nbats:
ty but we'll need to test ourselves first

PR #1319, opened 2024-01-05
Update VideoPiracyGuide.md

RRR-sudo:
Added new Site for Anime


nbats:
thank you we'll test it out and add if people like it

PR #1320, opened 2024-01-05
Update iOS section

theredspecial0:


nbats:
I was editing this page, I did my best to merge both together but if I missed any of your changes feel free to re-send them.

PR #1342, opened 2024-01-12
Update GamingPiracyGuide.md

DogancanYr:


nbats:
ty but we'll need to check it out ourselves first

DogancanYr:
It's like GreenLuma, but it has an interface and you can also make settings on an extra steam account. (Family pin Brute-force attack, stopping game updates, opening DLC, showing games you don't own as purchased in the steam interface. [different crack from Goldberg] etc.

nbats:
alright I can't merge but I'll get it added

DogancanYr:
> check it out ourselves

Why are these being checked? Is it because they may contain viruses or because they are not good enough to be worth adding?

(I mean, because I don't disclose them?)

Would it be easier if I disclosed their content?

nbats:
Yeah given the fact that its something users install, things like virustotal results and explanations as to why they should be added will increase the likelihood of me merging right then, rather than putting it in discord to check later.

DogancanYr:
> Yeah given the fact that its something users install, things like virustotal results and explanations as to why they should be added will increase the likelihood of me merging right then, rather than putting it in discord to check later.

https://github.com/pjy612/SteamManifestCache/releases
I think it's because it's closed source, Steam might close the vulnerability And when I look, they use the same infrastructure Also, I will make a guide about this software tomorrow, after all, it is difficult to read or understand with its very limited guides and translation. 👍

WiLuX-Source:
@DogancanYr hey there, If it is okay for you how can I contact you? I have questions regarding this tool.

DogancanYr:
> @DogancanYr hey there, If it is okay for you how can I contact you? I have questions regarding this tool.

@WiLuX-Source Discord:DogancanYr I'm not available right now, I can check later

PR #1343, opened 2024-01-12
Update GamingPiracyGuide.md

DogancanYr:


DogancanYr:
In the original version, there is no auto-generate file and the achievements are incorrect.

DogancanYr:
> In the original version, there is no auto-generate file and the achievements are incorrect.

Why ?

nbats:
sorry but it only has 18 stars so we'll need to at least read through it ourselves before adding

DogancanYr:
> sorry but it only has 18 stars so we'll need to at least read through it ourselves before adding
Original version was problematic, so a new update was made on cs.rin.ru by users (2 months ago) and this is its continuation and slightly improved version.

nbats:
Alright fair enough, I'll get it added

PR #1358, opened 2024-01-16
Update GamingPiracyGuide.md

DogancanYr:


nbats:
thanks I'll get the steam packer added, but I feel like the guide is a bit too straightforward to be worth adding

DogancanYr:
> thanks I'll get the steam packer added, but I feel like the guide is a bit too straightforward to be worth adding

I think so. It's best if I do it myself.

PR #1361, opened 2024-01-16
Added user agent switcher to firefox extensions - Internet-Tools.md

TF2Northstar:
Added user agent switcher to firefox extensions list, I also starred this addition due to its ability to allow users to bypass youtubes 5 second block on loading new videos. (When using chrome as the user agent, this block is gone.)


nbats:
ty but according to May these type of extensions can make you easier to track. Talk to her if you want more info

PR #1366, opened 2024-01-17
De-cluttered To-Do list Android apps 📝

klenndal:


klenndal:
List of changes:
Fixed missing reddit link
Removed ROMs tag from the-eye
Added "(Debrid Services)" to the File Leeches for easier searching
Moved Real-Debrid to File Leeches, it is a website not an app
Moved Debrid-Link to File Leeches
Added AllDebrid to File Leeches, it has a week free trial
Removed https://hdencode.org/, it is in the wrong section

nbats:
Looks good thank you. I had to make some descriptions of the To-Do apps shorter so it doesn't look so bulky, and I also kept HDencode where its at as it requires something like Real-Debrid to be able to download.

PR #1369, opened 2024-01-20
Update GamingPiracyGuide.md

DogancanYr:
Things started to get complicated.


nbats:
sorry but instead of this format could you just use the format we normally use, and maybe change the descriptions of each instead?

DogancanYr:
> sorry but instead of this format could you just use the format we normally use, and maybe change the descriptions of each instead?

I'm trying to do it, but things get complicated. It doesn't look good to write about each of them separately.
I can make a horizontal version of this, but I need to add 3 more horizontal rows.

nbats:
I may not be following exactly, but it just needs to be formatted the way we always format stuff. You can split each into their own line if needed.

PR #1371, opened 2024-01-20
Update File-Tools.md

DogancanYr:


nbats:
ty but this is supposed to be a / to match how we do it in the rest of the wiki

PR #1373, opened 2024-01-20
Update GamingPiracyGuide.md

DogancanYr:


DogancanYr:
They unlock cracked games and their dlcs, and they also unlock dlcs you own in legal copies of the game.
I didn't know what to write.
Maybe keep it that way.

PR #1376, opened 2024-01-21
Update Android-iOSGuide.md

XCE0:
remove geddit,as it's dead app.


erikk333:
It works tho

XCE0:
Long-term, it's better to direct user of FMHY to active apps instead of geddit as if it stopped working, it will not be fixed.

nbats:
Yeah I'm kinda mixed but I feel like its best to keep it while it still works as their aren't many good options

PR #1390, opened 2024-01-23
Update Internet-Tools.md

DogancanYr:
Untrustworthy sites are included by default, and trusted sites are not added.


nbats:
True but the site is customizable, people can add or remove anything they'd like, so I still think it has enough uses to keep.

PR #1406, opened 2024-01-30
Update NSFWPiracy.md

XCE0:
Remove outdated websites


nbats:
I dont really see how this is outdated, it has almost 500 sites and lists a lot of popular ones.

PR #1407, opened 2024-01-30
Update NSFWPiracy.md

XCE0:
Remove uneeded websites


nbats:
Can you give some sort of reasoning for removing these? They seem to work fine, so maybe bumping them down lower rather than removing would be better.

PR #1421, opened 2024-01-30
Update VideoPiracyGuide.md

XCE0:
remove dead link


nbats:
Thank you for all the pull requests, I appreciate the help, but keep in mind you dont need to send every edit individually if its on the same page. You can make multiple edits on Video Piracy for example, then send them all at once as one pull request.

PR #1440, opened 2024-01-30
Update Android-iOSGuide.md

XCE0:
Remove dead project


nbats:
downloads on the release page still work

XCE0:
their latest release is from 2022.

nbats:
their github says it was updated 4 months ago

PR #1444, opened 2024-01-31
Update AdblockVPNGuide.md

XCE0:
Remove Dead projects


nbats:
I removed the two firefox addons, but ClearURLs + Canvasblocker have both been updated pretty recently.

XCE0:
latest release of clearurls is from 2022.

latest release of CanvasBlocker is from april the previous year and the git does not seem to be active.

PR #1446, opened 2024-02-01
Update Android-iOSGuide.md

stuffed2:
added jailbreak news site under ios jailbreaking


nbats:
ty we'll check it out and add if people like it

PR #1458, opened 2024-02-02
Update Internet-Tools.md

LegendaryITA:
Proton Pass


nbats:
Got it starred thank you, but they have the extensions / apps linked on the site itself so I dont think we need all of those.

PR #1468, opened 2024-02-03
Update may-2023.md

XCE0:
update link


nbats:
got it switched to .net

PR #1479, opened 2024-02-04
Update LinuxGuide.md

XCE0:
dead link


nbats:
got it removed + kept description

PR #1481, opened 2024-02-04
Update AdblockVPNGuide.md

DogancanYr:


nbats:
ty but we'll need to check it out first

DogancanYr:
> ty but we'll need to check it out first

It is already a well known antivirus. I don't think there will be a problem.

DogancanYr:
What kind of check will it be?

nbats:
sorry but according to one of our mods bitdefender is anti-piracy and also gets many false positives.

DogancanYr:
> sorry but according to one of our mods bitdefender is anti-piracy and also gets many false positives.

I'm trying to add vpn here, not anti virus.

nbats:
I see but if they're anti-piracy I still dont think its a good idea to direct people to them

DogancanYr:
> I see but if they're anti-piracy I still dont think its a good idea to direct people to them

How other vpns work, this vpn works the same, there is no difference, and even torrenting works. (also free)

nbats:
okay fair enough we'll take a closer look

PR #1490, opened 2024-02-05
Update AI.md

henlee1:


nbats:
thank you but we'll need to try it out ourselves first and we'll get it added afterwards

PR #1492, opened 2024-02-05
Update Non-English.md

Rust1667:
tag PlayDede as it can only be accessed with IP address from Spain. Unstar because its not available from South America. But keep up because it could be the best site for Spain


nbats:
got it updated

PR #1519, opened 2024-02-09
Update GamingPiracyGuide.md

DogancanYr:
I think it was good.


nbats:
ty but I dont think any of this is needed, most games come with these installers

DogancanYr:
That's true, but when you download a game without an installer you don't know what software will be installed.

PR #1521, opened 2024-02-09
Update sept-2023.md

XCE0:
Update links


nbats:
ty but tasky is going to replace all of them

PR #1526, opened 2024-02-10
Update GamingPiracyGuide.md

mikkoyr:
Deleted broken link, added new browser gaming link


nbats:
removed the dead site but not sure if hahagames is worth adding

mikkoyr:
Added hahagames as it was modern with tons of games, including some unique ones I havent seen elsewhere. Some other sites in the same group seemed lower quality like xefer (unsecure site, 2 games total), yell0wsuit.page (only 76 games), gamekb (a lot of broken games) etc..

PR #1528, opened 2024-02-10
Update ReadingPiracyGuide.md

GuiltiaSinJ:
Added and starred server.elscione.com, a simple site mainly for downloading light novels (for now at least) in epub and high quality pdfs, it offers a wide variety of novels from official TL LNs, unofficial ones, to even webnovels, on top of all this it also has manga in cbz although the library is pretty small for now and music (mostly japanese, and again the library is still small) in FLAC It's still got room for growing as the owner promised update 

By the way given how there is different stuff being hosted I didn't know which to put it b ut since it seems to be more focused on LNs I decided to put it here


nbats:
thank you, I'll need to base64 this, but I'll get it added to reading in a bit

PR #1532, opened 2024-02-12
Update NSFWPiracy.md

zinklog2:
reorganized manga section (unstarred nhentai and 3hentai, starred imhentai and hitomi)

### DIFF
--- a/NSFWPiracy.md
+++ b/NSFWPiracy.md
@@ -358,24 +358,23 @@
 
 ## ▷ Hentai Manga / Comics
 
-* ⭐ **[NHentai](https://nhentai.net/)** - Hentai Manga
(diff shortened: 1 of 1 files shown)


nbats:
kept all the changes except unstarring 3hentai and nhentai as both of them have massive libraries, especially for one shots.

nbats:
alright I'll bump it down more, but they still have 18k pages of manga so I dont think it needs to be unstarred

nbats:
and 3hentai I use all the time to find very obscure stuff. I just did the math and they have 227,925 individual mangas

nbats:
I agree they should both be higher than nhentai, I just dont think nhentai is worth unstarring

PR #1534, opened 2024-02-12
Update File-Tools.md

DogancanYr:


nbats:
ty but its already added in that section

DogancanYr:
> ty but its already added in that section

Search box doesn't work properly..

PR #1539, opened 2024-02-13
Update File-Tools.md

cyflux:


DogancanYr:
resim

nbats:
I think its worth adding, just not starred giving the limits

DogancanYr:
> I think its worth adding, just not starred giving the limits

It says that if you upload up to 10 GB, You can download it within 48 hours.  If it exceeds 48 hours, you can download it with money and that file will be deleted after 90 days.

There are a lot of sites on the page where you can upload files of more than 20 GB and they are not deleted after 48 hours.

nbats:
True but we also have a lot that are like 2gb and only 7 days. The limits here are annoying but not so much so that I think the site has no use.

DogancanYr:
> True but we also have a lot that are like 2gb and only 7 days. The limits here are annoying but not so much so that I think the site has no use.

Didn't you have the top picture opened? It says that it will be deleted after 90 days.

nbats:
so? most of the sites do that in the file hosting section.

DogancanYr:
> oh you mean the tag, my bad, I got it updated to 90 hours instead of days

Now it began to make sense. 👍

PR #1545, opened 2024-02-14
Update DownloadPiracyGuide.md

DogancanYr:


nbats:
ty but we already have it in the index section of miscellaneous

DogancanYr:
> ty but we already have it in the index section of miscellaneous

I know I just want it to come to the place that is relevant to the subject.

PR #1546, opened 2024-02-14
Update STORAGE.md

DogancanYr:


nbats:
thank you but these are all spread throughout the wiki already

DogancanYr:
> thank you but these are all spread throughout the wiki already

I know, but they should also be together.

PR #1548, opened 2024-02-14
Remove adblock testers

ItsProfessional:
Adblock testers are a terrible way of measuring if your content blocker is good. See this tweet by the author of uBO: https://twitter.com/gorhill/status/1583581072197312512

Also, I capitalized some product names.

### DIFF
--- a/AdblockVPNGuide.md
+++ b/AdblockVPNGuide.md
@@ -6,7 +6,7 @@
 
 # ► Adblocking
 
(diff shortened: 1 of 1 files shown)


DogancanYr:
> Adblock testers are a terrible way of measuring if your content blocker is good. See this tweet by the author of uBO: https://twitter.com/gorhill/status/1583581072197312512
> 
> Also, I capitalized some product names.

Posted 1.5 years ago. The sites are updated at certain intervals.

nbats:
I'll remove the star, bump them down, and add a warning with this tweet about accuracy.

nbats:
I'll remove the star, bump them down, but I think removing entirely is a bit extreme, they could still have some use imo. That tweet is from 2022 like Dogancan said, so they could have improved since then.

PR #1553, opened 2024-02-16
Update System-Tools.md

lauralex:
Added AnyDesk as a remote desktop app


nbats:
thank you but this suffered a really bad breach recently so people dont trust it anymore

PR #1555, opened 2024-02-17
Update Android-iOSGuide.md

stuffed2:
edits to clearer show which devices are supported. A12+ refers to the chip, it supports A12 chip and later (XS and later as well as the ipads that have it.) arm64 is the architecture of A11 and lower chips. You could edit it to say A9-A11 or something but I'm not sure what the lowest supported chip is.


nbats:
got the changed add ty

PR #1557, opened 2024-02-17
Update STORAGE.md

DogancanYr:


nbats:
got it base64'd and added ty

PR #1560, opened 2024-02-18
Update VideoPiracyGuide.md

CashyOP:
Add fossatv


malslarks (approved):

nbats:
thanks but this seems to be a clone of binged so I'll add it with that

PR #1563, opened 2024-02-18
Update AdblockVPNGuide.md

jijirae:
Added "The Second Opinion", which is a suite of malware/virus scanners that I compiled. (Users can also choose to download each of them one-by-one).

Famously used by The PC Security Channel
https://www.youtube.com/@pcsecuritychannel


nbats:
Thank you looks good, the only issue is that is has norton listed which we have in our unsafe list for being owned by avast. If that gets removed we'll add this list.

PR #1564, opened 2024-02-18
Update UnsafeSites.md

fruini:
Listed minecraft mod reposts.


nbats:
Thank you but this list is only for things that are actually unsafe to the users computer. I'm not defending what they do, its just not enough to call them unsafe.

PR #1568, opened 2024-02-20
Tarnkappe little change

jbugel-lol:
(I dont know why the other commits didnt clear but those are already live)
Little change on Tarnkappe. Instead of + theres now a /


nbats:
ty got it fixed

PR #1572, opened 2024-02-20
Update AdblockVPNGuide.md

DogancanYr:
Sounds like a good idea.

### DIFF
--- a/AdblockVPNGuide.md
+++ b/AdblockVPNGuide.md
@@ -423,6 +423,7 @@
 
 ***
 
+* 🌐 **[Do you need vpn when downloading files from torrent ?](https://www.comparitech.com/blog/vpn-privacy/internet-censorship-map)**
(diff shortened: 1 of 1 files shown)


nbats:
its interesting but I'm not sure if its useful in terms of torrenting. It has a lot of places like the US listed as uncensored where you do in reality need a VPN.

DogancanYr:
> its interesting but I'm not sure if its useful in terms of torrenting. It has a lot of places like the US listed as uncensored where you do in reality need a VPN.

How about we create it ourselves on a site like Github but that allows easier editing ?
Since there are many countries, torrent users there can change their own and we get accurate information.

nbats:
If you wanna take it on feel free, but it would be a lot of work to make sure its accurate especially given each ISP responds to piracy differently regardless of their countries laws.

PR #1578, opened 2024-02-22
Added WizTorrent torrent client

drakonkat:
Added WizTorrent client, free and with some innovative feature like the WebShare feature


nbats:
nice thanks for sharing, we'll need to test it a bit but we'll get it added soon if people like it

PR #1584, opened 2024-02-23
Update GamingPiracyGuide.md

DogancanYr:
I know it will go to the Base64 thread, I guess it doesn't matter if I write here and not through discord. :b


nbats:
no its okay only copyrighted stuff needs to be base64

PR #1590, opened 2024-02-25
BetterDiscord banned plugins

mixon606:
Adds github repository with plugins banned by the official BetterDiscord plugin repo


DogancanYr:
Top url should be added (https://github.com/mov-ebx/BannedBetterDiscordPlugins)

PR #1644, opened 2024-03-13
Update GD Tools Section

the9655a:
Changes:
- Moved all the lists onto one line for neater organization.
- Starred GDDP, amazing tool to practice and improve at the game. (I use it a lot)
- Added spritesheet splitter. Useful for making texture packs.
- Added Save Explorer
- Added Font Generator


stuffed2:
I don't think it's a good idea to put all the lists into one line, the gd community loves making lists and there is an absolute ton of them, I was planning to add some more

the9655a:
> I don't think it's a good idea to put all the lists into one line, the gd community loves making lists and there is an absolute ton of them, I was planning to add some more

Let's not add any more then. Only keep the main ones

PR #1694, opened 2024-03-23
Update Text-Tools.md

DogancanYr:
## Types of changes

- [ ] Bad / Deleted sites removal
- [ ] Grammar / Markdown fixes 
- [x] Content addition (sites, projects, tools, etc.)
- [ ] New Wiki section

## Checklist:

- [x] I have read the [contribution guide](https://rentry.co/Contrib-Guide).
- [x] I have made sure to [search](https://raw.githubusercontent.com/fmhy/FMHYedit/main/single-page) before making any changes. 
- [x] My code follows the code style of this project.

### DIFF
--- a/Text-Tools.md
+++ b/Text-Tools.md
@@ -81,6 +81,8 @@
 * [Translate Shell](https://www.soimort.org/translate-shell/) - Translation CLI / [GitHub](https://github.com/soimort/translate-shell)
 * [OnlineDocTranslator](https://www.onlinedoctranslator.com/en/) - Document Translator
 * [LegaleseDecoder](https://legalesedecoder.com/free-legal-document-to-plain-english-translator/) - Legal Document Translator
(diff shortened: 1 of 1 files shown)


nbats:
Added the first, but looks like Eazy Po is for "Gettext" files only.

DogancanYr:
> Added the first, but looks like Eazy Po is for "Gettext" files only.

Yes, but I wrote it this way because I couldn't find many such software that would be useful for someone who wants to translate.
It does not only support one file type, it supports text files like the one in the picture on the website.

DogancanYr:
Also, matecat does not translate files automatically, it needs someone's help. It can be thought of as an online translation editor and only this one has a google api key that you can use for free. It might even get a star on the site because of this feature. It is even possible to translate the file by itself by pressing Alt+enter, but it is not automatic.

nbats:
alright I'll get Eazy Po added and the other description updated

PR #1773, opened 2024-04-13
Update TorrentPiracyGuide.md

DogancanYr:
Even though it doesn't have a proper category setting, it's good for indexing download sites.


DogancanYr:
Or should I put it in the download section? I was undecided .

PR #1783, opened 2024-04-14
Fixed admonitions

the9655a:
Changes:
- Fixed all admonitions (dont add a "\*" character before them)
- Fixed "J2team", caused weird spacing on the website and i think it was meant to be starred
- Renamed "Modded Minecraft Launchers" to just "Minecraft Launchers", since the stuff here can still be used as just an alternative launcher and without mods, so its wrong to title them as Modded only.
- Moved "Pojav" under Minecraft Launchers
- Changed Pojav link to the correct URL.

### DIFF
--- a/AI.md
+++ b/AI.md
@@ -4,7 +4,7 @@
 ***
 ***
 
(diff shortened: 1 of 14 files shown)


nbats:
the * doesnt break the markdown with these, but I did get the rest of the changes added

PR #1796, opened 2024-04-17
Bug fixes

the9655a:
Changes:
 - Fixed a storage redirect URL
 - Fixed wrong order of storage links under News
 - Moved OSINT out from under Indexes and to its own category, since its not an index or list of indexes, but a standalone section
 - Tweaked admonition links

### DIFF
--- a/MISCGuide.md
+++ b/MISCGuide.md
@@ -80,7 +80,7 @@
 
 ***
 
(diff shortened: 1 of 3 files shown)


nbats:
Looks good but I'm not too sure about moving OSINT into its own section, most of them are list of tools and sites

PR #1814, opened 2024-04-23
Debloated Bypass Paywalls Section

the9655a:
## Changes:

**Note**: All links were thoroughly tested with a variety of different paywalled articles. You can find the full list [here](https://rentry.co/vw6mqrko).

- Starred:
	- "Freedium" - Amazing site, works 100% of the time (except one instance[^1]), no ads, really fast load times, and is responsive so you can view articles on mobile devices.
	- "Archive.is" - Works surprisingly well, especially for particularly tricky sites, including Medium articles. It's also useful for rare instances when other solutions don't work. Archive.is can load article images, something Archive.org sometimes can't. it also loads archived pages a lot faster than Archive.org.
- Removed:
	- "12ft.io" - Very hit and miss, sites don't always work, and also has been repurposed since its last takedown to just remove popups, banners, and ads, which something like uBO would be better for. Some sites can work as a byproduct of this, but its not reliable enough.[^2]
	- "1ft.io" - Similar to "12ft.io", i.e. hardly works and isn't designed to bypass paywalls, only to remove ads, which again uBO is far better at.
	- "Paywall Bypass Index" - Outdated + FMHY has a better list now.
	- "medium-unlocker" - Still has unresolved issues from last year, the dev also recommends that you use a pre-release to have a good chance of articles being bypassed. Considering its a beta build, and with seemingly no active development in the last 9 months, the extension might not work all the time, be very buggy, or likely have many other problems like [compatibility](https://github.com/und3fined/medium-unlocker/issues/55) issues.
	- "Scribe" - Broken.
	- "RemovePaywall" - Hardly works for normal articles, doesn't work on medium articles, and is not reliable enough to keep.
	- "Archive Buttons" and "PaywallHub" - Useless.[^3]
	- "PressReader" - Broken + 6 years outdated.
	- "Smry.ai" - Barely works + Has a fake supported sites list.[^4]
	- "bypass-paywalls" - Only one or two sites worked when tested, also isn't as good as BPC or as powerful, with not as many supported sites either.
	- "shacklefree" - Only worked twice during my tests and sometimes loads articles with no CSS, so the viewing experience is awful.
	- "Readium" - Doesn't work + No updates in 4 years.
- Untested but still removed:
	- "PaywallBypasser" - No updates in a year, only 35 stars, doesn't seem useful.
	- "Hover Paywalls" - Last release 4 years ago, no active development, so likely doesn't work today.
- Other:
	- Fixed a redirect link for the section.
	- Moved Mozilla's Pocket to a more appropriate place.
	- Tidied up existing links.

[^1]: All the articles I tested worked perfectly, except [one](https://freedium.cfd/https://levelup.gitconnected.com/introducing-pandasai-the-generative-ai-python-library-568a971af014), which showed only a blank page with no article. All other articles I tested from the same domain (levelup.gitconnected.com) worked fine, so its highly likely this is **just a bug** and shouldn't affect whether it's starred or not, as it still works flawlessly and is super convenient. In the meantime, I've raised [an issue](https://codeberg.org/Freedium-cfd/web/issues/4) on the sites repository.
[^2]: Moved it under adblocking instead, since it works for simply sharing a cleaner version of a URL, which some might find useful.
[^3]: Both sites have buttons for the following: "archive.is", "archive.org", "removepaywall.com", "12ft.io", and "google cache", 4 of which are either useless or bad: Google cache was discontinued a while ago and no longer exists; "removepaywall.com" and "12ft.io" have been removed for reasons stated above; and "archive.org" is slow and doesn't always work. It also can't bypass Medium article paywalls; "archive.is" can. That leaves only "archive.is", which can just be linked separately. Additionally, the BPC extension [actually grabs content from "archive.is"](https://i.ibb.co/Ms31MLT/img.png) for certain articles, so sometimes you won't even need to use the archive site in a separate tab. It is for all these reasons that these 'buttons' are completely useless.
[^4]: It could go with "12ft.io" under adblocking because its similar, but I'm not sure how popular 'sharing clean article links' is. It seems pretty niche, so I kept it removed.

### DIFF
--- a/AdblockVPNGuide.md
+++ b/AdblockVPNGuide.md
@@ -22,6 +22,7 @@
 * [Popupblocker All](https://addons.mozilla.org/en-US/firefox/addon/popupblockerall/), [PopUpOFF](https://romanisthere.github.io/PopUpOFF-Website/index.html) or [PopupBlocker](https://github.com/AdguardTeam/PopupBlocker) - Popup / New Tab Blockers
 * [Adguard](https://github.com/AdguardTeam/AdguardBrowserExtension#installation) - Adblocker
 * [Adblock Tester](https://adblock-tester.com/) or [AdBlocker Test](https://d3ward.github.io/toolz/adblock.html) - Adblocking Tests
+* [12ft.io](https://12ft.io/) - Share Ad-Free URLs
 
 ***
 
--- a/Internet-Tools.md
+++ b/Internet-Tools.md
@@ -126,6 +126,7 @@
 
 * [Bypass paywalls for scientific documents](https://greasyfork.org/en/scripts/35521) - Bypass Scientific Document Paywalls
 * [unpaywall](https://unpaywall.org/) - Bypass Scholarly Article Paywalls
+
 * [PressReader](https://greasyfork.org/en/scripts/39936) - Bypass PressReader Paywall
 * [Burles](https://burles.co/en/) - Paywall Bypass
 * [bypass-paywalls](https://github.com/iamadamdev/bypass-paywalls-chrome) - Paywall Bypass
@@ -136,6 +137,8 @@
 * [Smry.ai](https://www.smry.ai/) - Paywall Bypass
 * [Readium](https://sugoidesune.github.io/readium/) - Paywall Bypass
 * [shacklefree](https://www.shacklefree.in/) - Paywall Bypass
+
+* [Archive.is](https://archive.is/) or [Archive.org](https://archive.org/) - Archived Articles
 
 ***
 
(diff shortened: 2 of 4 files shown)


lenicyl:
> "Scribe" - Broken + Links are tedious to create

What exactly is broken or tedious about scribe.rip ? It seems to be working fine for me

the9655a:
> What exactly is broken or tedious about scribe.rip ? It seems to be working fine for me

All of the medium articles I tested didn't work with the site, so it is presumed broken. There are also way better alternatives, like Freedium.

radiohead-sucks:
just chiming in to say that I vouch for freedium, it does in fact deserve a star

nbats:
Looks good, wont let me merge due to conflicts, but I did add all the changes.

PR #1830, opened 2024-04-26
Update GamingPiracyGuide.md

stuffed2:
added a site and improved the description of another. The site seems a bit dodgy but I tried it (with ublock and antivirus) and it actually works so it seems very useful for switch emulation setup


stuffed2:
By the way I would consider starring emudeck, it is very seamless and useful and it now supports windows, android, and more linux distros as well as steamOS. Don't put gameimage with it, it is not relevant

nbats:
looks good but I'm not sure its a good idea to add the key site

PR #1848, opened 2024-04-30
Update GamingPiracyGuide.md

DogancanYr:


nbats:
We'll need to look through it a bit but I'll get it added after we do

PR #1851, opened 2024-05-01
Update Non-English.md

Rust1667:
add spanish DDL site


Rust1667:
Im sure
But the link shortener requires either the bypass userscript, or disabling ublock origin, or using adguard

Rust1667:
that link shortener "mega-enlace" is also present in many of the spanish DDL sites already in the wiki

nbats:
I have the userscript on and there was nothing seamless for me, it took me to like 5 pages and even to google. There's just no excuse for that.

Rust1667:
Idk, Im with the userscript and its: a page (where you wait 30s), which leads to a pastebin. The pastebin has shorteners that are instabypassed to the file host.
I added one of those bypasses like yesterday to the userscript, might need to upadte it.

nbats:
the issue is that most people wont have this script, so we're basically just sending a majority of people to a bad site

PR #1859, opened 2024-05-02
Update game-tools.md

XtronXI:
I think MCBEDOC shud be added as star, but i wont do it. Its your wish if you wanna.


nbats:
thank you I got it added to base64 to be safe and I'll link that in the main section

PR #1877, opened 2024-05-05
remove the space in gamebounty (again)

radiohead-sucks:
why both those words had been split again is beyond me, they literally are always together, it makes zero sense to add a space between them


radiohead-sucks:
:lowcredit:

zdashero:
Just to add to this we have the words split on Google for SEO purposes GameBounty is the official spelling

PR #1910, opened 2024-05-15
Added OfflineMinecraftLauncher to Minecraft Launchers

antunnitraj:


nbats:
Thank you but we'll need to try it first ourselves

PR #1940, opened 2024-05-25
Update File-Tools.md

FMHY-Froggy:
Removed anondrop at the owners request.
See #to-do for context. (I have verified that the account which requested the removal is indeed the owner)


nbats:
thank you its gone

PR #1958, opened 2024-05-31
Update

the9655a:
- Removed "[4GB Patch](https://ntcore.com/?page_id=371)" from under Optimization Tools - it's niche. Not a lot of games are likely still using 32bit, otherwise I feel this tool would be more popular than it is. It's only really mentioned in the GD space, as it allows you to load large levels and prevent crashes. When GD (hopefully soon) moves to 64bit, it can be removed from the GD section.
- Moved:
	- "Patch Reddit Clients" under Reddit Tools.
	- "Eaglercraft" note to the front.


nbats:
The app patch thing is for mobile apps so I'm not sure why we'd move that, but I did get the other changes added.

the9655a:
Yeah mb. Ig I felt it would also be cool if it was discoverable from there too, but yes it fits better on the android page.

PR #1961, opened 2024-06-01
Update VideoPiracyGuide.md

Im2rnado:
Added faragski.com site


nbats:
thanks but looks like its just a vidsrc.pro site and we already have ones that have both this and other on the same site.

PR #2032, opened 2024-06-19
Update AdblockVPNGuide.md

the9655a:
- Improved tip messages regarding adblocking on chrome.


nbats:
I think it would be best just to link it normally, there's no way the average person is going to read a note that long.

nbats:
I just added ubo lite as a entry, I think its way more likely people see it that way.

nbats:
My bad just got it it fixed

PR #2071, opened 2024-06-29
Update DEVTools.md

stuffed18:
I don't know what your policy is on adding youtube videos, but this one was a genuinely useful guide that helped me get into decompiling games and explained it well, I usually find most people like to say its not possible or give vague advice when googling this topic


nbats:
Looks interesting but I'm not sure if most people would have a need for it

PR #2079, opened 2024-07-01
added HDencode

Vojtak42:


nbats:
thank you but we already have it posted next to real-debrid, I think you need RD to use it

PR #2103, opened 2024-07-07
added youplex

steveyout:
## Types of changes

- [ ] Bad / Deleted sites removal
- [ ] Grammar / Markdown fixes 
- [ ] Content addition (sites, projects, tools, etc.)
- [ ] New Wiki section

## Checklist:

- [ ] I have read the [contribution guide](https://rentry.co/Contrib-Guide).
- [ ] I have made sure to [search](https://raw.githubusercontent.com/fmhy/FMHYedit/main/single-page) before making any changes. 
- [ ] My code follows the code style of this project.


nbats:
thank you but this was removed in the past as its only 2 servers which most sites already have.

steveyout:
but cant it still be include in the wiki??

nbats:
not really, there are hundreds if not thousands of vidsrc / uplcloud only sites, so if we allow one we have to allow them all. It needs to stand out more in some way, more servers, self-hosted videos, something like that.

PR #2110, opened 2024-07-09
Update AI.md

kushagra-xo:
- adds venice ai, a generative app that is open, private and uncensored.
- restructure headings with respect to importance

feel free to revert the restrucsturing.

### DIFF
--- a/AI.md
+++ b/AI.md
@@ -4,30 +4,6 @@
 ***
 ***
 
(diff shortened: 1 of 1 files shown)


nbats:
Thank you, we'll take a look at Venice, and I'm not sure about the section re-ordering but I will consider it.

PR #2175, opened 2024-08-02
Update VideoPiracyGuide.md

crossdotmoe:
removed dead vidsrc.to api, added mp4hydra.org and 90min.to


nbats:
Thank you I'll get the new sites added to testing to see what people think.

Things I did:
- Bumped down susflix.
- Bumped up freetubespot / 9PM.
- Bumped down flixtor.
- Bumped up animez / animension.

Things I didn't do: 
- I kept Braflix where it was, it did better than fbox in our polls. 
- I kept the use adblock label on upmovies, it has lots of popups for me.
- I kept vidsrc for now since I think it'll come back, but it doesnt soon we'll remove.

PR #2190, opened 2024-08-05
Update non-english (Indonesian)

slashedzer0:
Fix grammar and small correction


nbats:
I got the Indonesian one updated, but you removed a bunch of the other ones without explaining.

PR #2210, opened 2024-08-11
Update devtools.md

stuffed18:
so I saw this here, you normally have to pay like $20 for pico-8, but the download files are actually hidden on the site and u can download them without owning it

I put them into the pastebin


nbats:
ty I appreciate it but in general we dont direct link stuff like this

PR #2244, opened 2024-08-18
Update file-tools.md Image to PDF

stuffed18:
Added image to pdf tool. I really like this because I use it often to convert comics to pdf. It works really fast, I just did a 250 png conversion to pdf and it took like 20 seconds. You don't need to download any tools or upload it to any server, it works completely offline.


nbats:
got it added ty

PR #2348, opened 2024-09-08
Update social-media-tools.md

itsvijaysingh:
add discord pfp viewer tool, lookup tool and collection website


Setsuwaa:
Why did you add it twice? Also, there's two links for the same thing right above that.
pfpfinder.com isn't meant just for DLing pfps but other things too, so I think it should be added to img-tools#icons-avatars or /social-media-tools#client-customization. (It does advertise itself as a _discord_ pfp tool, but I don't see any reason to restrict images to one site when they can be used everywhere which is why I don't think it should be added to client-customization.)

itsvijaysingh:
> Why did you add it twice? Also, there's two links for the same thing right above that. pfpfinder.com isn't meant just for DLing pfps but other things too, so I think it should be added to /img-tools#icons-avatars or /social-media-tools#client-customization. (It does advertise itself as a _discord_ pfp tool, but I don't see any reason to restrict images to one site when they can be used everywhere which is why I don't think it should be added to client-customization.)

Hi ,

I've made the following updates based on your feedback:

> Removed Duplicates: The redundant entries for site, in social-media-tools.md have been removed.

> Re-categorized Content: I’ve added the entry to the appropriate section in img-tools.md.

Setsuwaa:
Looks good!

PR #2442, opened 2024-09-30
Update videopiracyguide.md

crossdotmoe:
added booflix.to, made by admin of meowatch.tv
self hosted video library
Still in development


nbats:
ty we'll check it out when we can

PR #2449, opened 2024-10-03
Add teeworlds and ddnet oss games

ChillerDragon:
Could also be added to macOS and linux games because both are cross platform having full native windows, macOS and linux support. DDNet even runs on android.

The descriptions come from their github:
- https://github.com/teeworlds/teeworlds
- https://github.com/ddnet/ddnet


Setsuwaa (approved):

nbats:
Thank you but we don't really add downloadable games for the most part.

ChillerDragon:
The list i added it to is full of downloadable games?

nbats:
yeah but they're open-source versions of really popular things like roller coaster tycoon for instance.

PR #2501, opened 2024-10-17
Hyperlinking pages in feedback messages

god-damnit:
Made it so that feedback messages have the "Page" hyperlinked. 
~I hope it will work~


taskylizard (approved):
I'm not fond of this change. However, I'll still merge this in until I work on this in the future. Wotaku already has [this](https://github.com/wotakumoe/wotaku/blob/c1f7ca4e5df9dfb2f9ea2c2e386b0e67c504dbac/docs/.vitepress/theme/components/Feedback.vue#L19) but it's an undesirable solution.

PR #2814, opened 2025-01-07
starred 8bitdash

beastbond:
I checked 8bitdash out and I was amazed by how good looking this startpage is, I wondered why wasn't it starred so here is it. This is my first contribution thanks for your handy guide


nbats:
thanks for the suggestions, we'll take a look at it and make a decision

PR #2845, opened 2025-01-14
Update miscguide.md

timboxx:
Added Geizhals to the shopping list.


nbats:

PR #2854, opened 2025-01-16
Update videopiracyguide.md

mepraz:
I have added Cinebook, because they've got many features like a whole social media for cinephiles, movies according to mood, multiple servers and many more features.


nbats:
I will give a extra point for the mood based search

PR #2856, opened 2025-01-18
Added Zoom Earth

felixs-alt:
Great website for live satellite images of the earth every 10 minutes


nbats:
ty but we already have this under Climate / Weather

PR #2864, opened 2025-01-20
Add ImgUpscaler AI

HelloVan123:


nbats:

PR #2866, opened 2025-01-20
Update adblockvpnguide.md

Jji666:


nbats:
ty I'll get both things reported remove in the next update

PR #2870, opened 2025-01-22
Update readingpiracyguide.md

NomadVibes88:
Reading CSE: Fixed Formatting.

Under `Ebooks`, Reading CSE is listed as 'Reading CSE**', which isn't the correct formatting. This is why I've fixed the formatting by removing the two stars.


nbats:
ty its not letting merge but I will get it fixed

PR #2880, opened 2025-01-24
Add glowing star emoji to emoji legend

god-damnit:
Added the new glowing star emoji for most recommended sites in emoji legend
image


nbats:
I dont think this is really needed, we only do it for 1 section and its pretty obvious what it is

PR #2886, opened 2025-01-25
Update non-english.md

Ayanokoji42069:
Added Sport section in "Czech / Čeština" that are free to watch


nbats:
thanks for sharing, we'll check them both out when we can

PR #2900, opened 2025-01-28
Add GoSearch

ibnaleem:


nbats:
sorry but we dont add tools that could be used for malicious purposes

PR #2919, opened 2025-02-02
Update readingpiracyguide.md

KinGsNowGH:
Added a h2 topic about book summaries, really interesting, couldnt find anything like this anywhere, i did some researches on reddits and personal choices to find all theses sites.


nbats:
thank you we'll take a look through them all when we can

PR #2921, opened 2025-02-03
Update non-english.md ( Added UHD Movies - known for 4k movies in eng…

gamersroyal042:
…lish and hindi dub )


nbats:
thank you I'll get it added, but we have another link that auto-redirects to the new site, so I'll use that instead

PR #2928, opened 2025-02-04
Fix formatting on Knaben

JourneyJ012:
added an open bracket for the URL to fix it.


nbats:
ty I'm editing right now so I'll get it fixed

PR #2929, opened 2025-02-04
Update android-iosguide.md

Rust1667:
add notes to Android movie apps about the required players.

This got me confused and led me to think the apps didn't work. The modded version of those required players are include in the mobilism posts, but I think other people could miss them too.

This is similar to that warning note added next to HDO Box, which also requires a player (Drama Player) for Android TV


nbats:
looks good, I'm editing right now so I'll apply these soon

PR #2930, opened 2025-02-05
Adding a Telegram Channel into games

0xCurtis:
My friend has a nice telegram where he share crack for games he likes (sometimes also anime episodes but mainly games)


nbats:
thanks for sharing we'll take a look when we can

PR #2932, opened 2025-02-05
Update videopiracyguide.md

Rust1667:
adding https://acestreamui.app/
this web-app requires to have Ace Stream running in the same device to work. (I tested) It is really good because it auto-detects all the available channels in the acestream network, filter or search them, and play them

### DIFF
--- a/docs/videopiracyguide.md
+++ b/docs/videopiracyguide.md
@@ -724,7 +724,7 @@
 * [Autobrr](https://github.com/autobrr/autobrr) - Autodownload Movie Torrents
 * [Powder](https://powder.media/) - Torrent Streaming App
 * [Popcorn Time](https://popcorn-time.site/) - Torrent Streaming App / [GitHub](https://github.com/popcorn-official/popcorn-desktop/)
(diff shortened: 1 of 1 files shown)


InternetNinjo:
> adding https://acestreamui.app/ this web-app requires to have Ace Stream running in the same device to work. (I tested) It is really good because it auto-detects all the available channels in the acestream network, filter or search them, and play them

this doesn't look safe as it needs users to install unknown extensions

Rust1667:
I used it without that chrome extension. Just installed Ace Stream.
The extension apparently is to be able to watch the stream on the web-app.
But the main utility is not that, because you can watch the streams in the Ace Stream native player.
The utility is to search for active streams through the active network.
Anyway the chrome extension they suggest https://chromewebstore.google.com/detail/easy-cors/gcdaaelgdlicnnichhholnoagafangej doesn't look too bad but idk.

nbats:
doesn't work for me, just shows a blank green page if you refresh once, I tried on FF and chrome.

PR #2937, opened 2025-02-05
Update internet-tools.md

magicoflolis:
Added my userscript which finds available UserScripts for the current webpage.
https://github.com/magicoflolis/Userscript-Plus


InternetNinjo:
> Added my userscript which finds available UserScripts for the current webpage. https://github.com/magicoflolis/Userscript-Plus

could be somewhat useful, but testing needed

PR #2941, opened 2025-02-06
Update videopiracyguide.md

qaisarshahzad1:
Hello Sir Add This Site


InternetNinjo:
> Hello Sir Add This Site
@qaisarshahzad1 
it could be better to join discord and add there instead, site seems to work, haven't tested streams

PR #2944, opened 2025-02-06
Update non-english.md

Adriann25:
Added mycomics. Unlike ComicMafia, you can read online, has no download feature.


nbats:
we'll take a look at it soon ty for sharing

PR #2953, opened 2025-02-08
Update edupiracyguide.md

ClassyD:
Free tools for music producers and songwriters, including a chord progression generator.


nbats:

PR #2961, opened 2025-02-09
Update edupiracyguide.md

NamelessGO:
Added Ultrasound / POCUS entry

Asking a second opinion on sorting the most relevant to least


nbats:
thanks for sharing, we'll take a look through the sites soon

PR #2971, opened 2025-02-12
Update non-english.md

sportontv:
Added a new live sports website


nbats:
ty we'll check it out soon

PR #2981, opened 2025-02-14
Update system-tools.md

memerman420:
Recently released Win 11 Debloater

Simple to use and open source


nbats:

PR #2983, opened 2025-02-14
Update ai.md, storage.md

shawnlee000:


nbats:
ty for sharing we'll check them out when we can

PR #2992, opened 2025-02-16
Update downloadpiracyguide.md

officialkaizen:
Added rlsdb to Leeches/Debrid section.
Not sure if "DDL Forums" is the correct term for the two sites, but couldn't think of any other fitting description for the sites.


nbats:
not letting me merge but I will get it added ty for sharing

PR #2994, opened 2025-02-17
Update non-english.md

gogoanimex:
we add Zorox a good ( Arabic anime streaming site)


nbats:
Yeah it will be added in the next update ty for sharing

PR #3009, opened 2025-02-20
Update text-tools.md

QuLi-ops:
This free tool can be used for character creation, including but not limited to naming anime characters, game characters, and novel characters. It can also be used for creating Japanese nicknames.


Setsuwaa (approved):

nbats:

Issue #3012, opened 2025-02-20
Movies Website


Rumam786:
### Type

Site suggestion

### Add additional context

Can you please add my movies streaming site.
This is my website link
          👇
https://fmhy.to

Setsuwaa:
is it just me or is this fmovies

nbats:
This has nothing to do with us

PR #3013, opened 2025-02-20
Add Fmhy.to to videopiracyguide

Setsuwaa:
Solves #3012 
Second commit was just me capitalizing the "F" btw


nbats:
this has nothing to do with us + its very slow

PR #3023, opened 2025-02-22
Update internet-tools.md

DoThingsWithAI:
- add a URL tool


DoThingsWithAI:
Ok, Thank you for pointing this out! I've added the required spaces before dashes throughout the code.
Let me know if there's anything else I need to adjust! 😊

nbats:

PR #3029, opened 2025-02-23
Update readingpiracyguide.md

Dinopirate06:
Added Readgraphicnovels.blogspot, it's a small, but high-quality library for award-winning graphic novels. Main purpose is to avoid censorship in countries where these novels are banned, but can be used for :pirate_flag: as well


nbats:

PR #3048, opened 2025-02-26
add findcryptojobs.com

goodwilliam:


nbats:
thanks for sharing we'll take a look when we can

PR #3049, opened 2025-02-26
Update storage.md

DogancanYr:
Steamworkshopdownloader.ru and ggntw are the same site.

Skymods is promising.

Also how to download workshop items from Ascendara?

### DIFF
--- a/docs/storage.md
+++ b/docs/storage.md
@@ -961,7 +961,7 @@
 
 ### Websites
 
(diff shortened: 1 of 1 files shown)


nbats:
ty looks good, and Ascendara has one if you scroll down their homepage iirc

DogancanYr:
I can't download the workshop from Ascendara, nothing happens.
Does it only support anonymous downloading?

nbats:
I'm not sure but I'll remove it from that section

PR #3051, opened 2025-02-27
naming consistency between index.md and wiki

wkaisertexas:
making each card name correspond to the title of the respective page of links

when using the site. the fact that the names of the cards are not an exact to the titles in the sidebar often gets me confused.

of course I know that `Movies / TV / Anime` *probably* corresponds to `Streaming`, but it would be better if the user did not have to think about it.

---

given the updates to these title cards, re-writing the descriptions could be in order. however, that can be an opinionating decision, so i will leave to the maintainers


nbats:
thank you but I think in this case its best to keep the titles short, it just looks better on the frontpage in my opinion.

wkaisertexas:
It's annoying for people who frequently visit the site because the titles are not an exact match. It's only a bit more text. The extra characters are well worth it in my opinion

If you are really concerned about text size, I would edit the wiki page to be 'Streaming' instead of 'Movies / TV / Anime'. However, I thought 'Movies / TV / Anime' was more descriptive so that is why I thought the homepage should be edited instead

PR #3060, opened 2025-03-01
Update miscguide.md - add Nomadlio.com

yaircohendev:
Hello I'm Yair the founder of Nomadlio, would love to add it to the wiki.

It's an alternative to Nomads.com (previously Nomadlist), I recommend adding it because it has a few big advantages to Nomads.com:
- It is 100% free, unlike Nomads.com which is paywalled after starting to use the website
- We're fully transparent about where the data is from in the website and FAQ
- It has ~5600 destinations, 4x more than Nomads.com


nbats:

PR #3063, opened 2025-03-02
Sonic Unleashed PC Port

iLusion3431:


nbats:
sorry it is a cool project but I feel like it could be a bit risky to post

MIOGMIOG:
> sorry it is a cool project but I feel like it could be a bit risky to post

it requires an actual game dump, but yeah you're kinda right I guess

nbats:
yeah you're right it doesn't have any game assets itself, probably no issue adding it then 

image

PR #3068, opened 2025-03-02
Update miscguide.md

alonelysaber:
Removed creddle dot io from Resumes section. Creddle was shut down in december 2024.


nbats:
I'll get it removed, ty for the heads up

PR #3082, opened 2025-03-05
Add Dutch sites - Update non-english.md

PcLover2:
This PR adds a lot of Dutch sites


nbats:
looks good ty for sharing, we'll need to check them out ourselves first, but we will soon

PcLover2:
Has this been looked at yet?

nbats:
yeah it has but we weren't sure on all of them:

* Duken - we generally dont add non eng indexes just because its too tough to confirm if they list good sources
* eboek - I'll get eboek added right now
* GP2B - What is this discord for exactly? We need to know what type of content it has + what it hosts are.
* puzo  - Hard for me to tell, I don't have an account so not exactly sure what it is, if you can send a screenshot we may consider it
* Torrentheaven - we did add this

PR #3101, opened 2025-03-07
Update non-english.md

DogancanYr:
GeoBlocked


nbats:
thanks for sharing we'll test them soon

DogancanYr:
These sites were already added to the site, they were just deleted because they were geoblocked.

nbats:
fair enough I'll get them re-added in the next update

PR #3112, opened 2025-03-09
Update readingpiracyguide.md

DanFMHY:
Tastedive links to games, but not movies or books


nbats:
I see /books now so must've been changes already

PR #3113, opened 2025-03-09
Update videopiracyguide.md

DanFMHY:
Tastedive links to games, but not movies or books


nbats:
alright I think this did get changed at some point, it says /books now

for the movie one front page is fine since it works for shows + movies

PR #3114, opened 2025-03-09
Update img-tools.md

n4zi-cipher:
added 'cyberdrop-dl' which allows downloading images from forums like F95zone


nbats:
ty for sharing, we'll test it soon

PR #3126, opened 2025-03-12
Update adblockvpnguide.md

DogancanYr:
Redundant with uBlock Origin's [removeparam](https://github.com/gorhill/uBlock/wiki/Static-filter-syntax#removeparam) and added lists. Any potential extra coverage provided by additional extensions is going to be minimal
https://github.com/arkenfox/user.js/wiki/4.1-Extensions#-dont-bother


nbats:
Our understanding is this does have a few things the list itself doesn't, we've considered removing it in the past and it was defended by multuple people

PR #3127, opened 2025-03-13
Add FLUX AI ART

ryancatter:


nbats:
thanks we'll check it out soon

PR #3133, opened 2025-03-14
Add Kotatsu to the Manga guide

RockinProjex:
I love this app


nbats:
thank you but we have it in the android section already

RockinProjex:
I know but I thought it would make sense to add it here too.

PR #3142, opened 2025-03-16
removed flicky stream

FifthWit:
flickystream is giving 500 errors. unusable


nbats:
yeah they're making upgrades right now but should be temporary

PR #3159, opened 2025-03-19
Replace WisprFlow with VoiceInk as a audio transcription software

j-jacobi:
VoiceInk is a early open source project that does audio transcription with a bunch of extra powerful features.

More: https://github.com/Beingpax/VoiceInk or https://tryvoiceink.com/

Wispr Flow has a subscription model and the free tier is quite small so I suggest removing it from this list.


nbats:
thanks for the heads up, we'll check out the new one, and review WisprFlow when we can

PR #3161, opened 2025-03-20
Replace MyFlixed Domain

dupitydumb:
Replace MyFlixed to new domain, the previous one is suspended


nbats:
we'll get it fixed in the next update, thanks for the heads up

PR #3165, opened 2025-03-20
Update devtools.md

sandipb2025:
Added Jsonguide's online json formatter tool


nbats:
thanks for sharing, we'll take a look at both when we can

PR #3181, opened 2025-03-24
Added https://easybreath.work/

ddsky:
free breath work app


nbats:
thanks we'll check both out when we can

PR #3182, opened 2025-03-24
Added Zen Browser

flaaaaaaaaa:
Open Source privacy focused browser based on Firefox


nbats:
Thank you but in general we dont add normal browsers, we have index of them (that includes zen) listed at the top of the browser section in internet.

PR #3183, opened 2025-03-25
Update img-tools.md

davidrey00:
New tool - image converter


nbats:
thanks we'll take a look when we can

PR #3186, opened 2025-03-26
Add Receipt Faker

erikasmalisauskas:


nbats:
thanks for sharing but not sure it really fits anywhere

PR #3197, opened 2025-03-29
Added Youtube Shorts Downloader

Balram-Parmar:
Modified Two files and added https://easyfetch.me/ which is fully working shorts downloader online


nbats:
thanks for sharing, we'll take a look soon

PR #3203, opened 2025-03-30
Update videopiracyguide.md

dht20:
Added 2 live sport streams


nbats:
ty we'll check both out when we can

PR #3208, opened 2025-04-01
Update text-tools.md

666de6:
add Bold Text Generator


nbats:

PR #3210, opened 2025-04-01
Update devtools.md: add svg convert tool

sing1ee:
Add SVG Converter tool, convert svg code to PNG/WEBP/JEPG/ICO.


nbats:

PR #3218, opened 2025-04-02
Add LootXP

ldobreira:


nbats:

PR #3221, opened 2025-04-03
Move P-Stream to properly differntiate from older & unmaintained movie-web sites

Xcissors280:
P-Stream is the only real maintained movie-web fork and thus should be included separately, it also offers other sources like the exclusive 4k token Fed-API which the wiki guide explains


nbats:
I dont know what that means really? The page lists instances of movie-web, not sure why we'd list this random one separately.

PR #3228, opened 2025-04-04
Rename codeium to windsurf

Indithem:
I'm not sure why this isn't been done, but I've renamed to include windsurf and also changed the link. The earlier link simply redirects to the newer one.

Also is this still starred? I'm no expert on that website..


nbats:
thanks for the heads up, we'll get it updated

PR #3230, opened 2025-04-04
Update videopiracyguide.md

dht20:
Fixed a link
and added 2 sites


nbats:
we already have topsport, but I will get the daddylive link fixed and we'll check out the other

PR #3233, opened 2025-04-04
Update linuxguide.md

hhoneeyy:
removed bluecomet as it's outdated and doesn't work anymore and added and starred an alternative (im pretty sure its a fork of it) SLSsteam


nbats:
thank you I'll get blue removed and we'll try the other

PR #3243, opened 2025-04-07
Update non-english.md

terrorbladegodlike:


nbats:
thanks but we dont add recipe sites given how many exist

PR #3244, opened 2025-04-07
Update non-english.md

Adriann25:
Added [beidian](https://www.beidiannet.com) and [myndfulmovers](https://www.myndfulmovers.com) for Chinese. They both have a lot of movie and cartoon dubs, which are fairly hard to find. Re-added [ComicMafia](https://comicmafia.to) for German. Not sure how it got removed, could not find any information on it.


nbats:
thanks for sharing, I'll get the label updated, and we'll check out the 3 new ones soon

Adriann25:
Unfortunately, the two Chinese websites have been shut down or are no longer accessible outside of China. Please do re-add the German one whenever you have the time though.

PR #3247, opened 2025-04-08
Added P-Stream individually as its the only active fork of movie web.

Xcissors280:
P-Stream is already #1 in movie web instances but most of the others are pretty outdated so it should be added individually and that section should be considered for removal.

it is open source and based on movie web and meets all the same criteria/rating, it does have minimal non intrusive ads.

### DIFF
--- a/docs/videopiracyguide.md
+++ b/docs/videopiracyguide.md
@@ -10,6 +10,8 @@
 
 ***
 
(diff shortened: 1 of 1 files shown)


nbats:
I really dont see the point, its #1 on the instance page already, why would we post it twice? They get more than enough attention being on the top of our list and movie-web.

PR #3261, opened 2025-04-10
Update text-tools.md

oussama0012:
new online notepad , super fast with unique options like auto save every 5 second , extract to many file .... and more


nbats:

PR #3264, opened 2025-04-10
Update links for willow

monyasau:


nbats:
looks good, sorry it said 720p there for so long in the label, I forgot to remove it, but I will fix it now + add the discord

PR #3283, opened 2025-04-13
good

007zzzz:


nbats:
Yeah I see your point, I'll change what you did for guides + communities and bump them up since they're useful for beginners, and I'll move the system stuff under apps since that's what they are

PR #3298, opened 2025-04-17
Update audiopiracyguide.md

Icyvert:
fixed a name and added uppercase where missing (verified from the respective sites)


nbats:
got it fixed thank you

PR #3309, opened 2025-04-20
Add hungarian streaming websites + update a redirect

misike12:


nbats:
ty got the broken one fixed + we'll check out the others

PR #3311, opened 2025-04-21
Update gamingpiracyguide.md

DogancanYr:
ShadPS4 emulator does not share the PS4 system module and is quite difficult to find online.


nbats:
thanks for sharing, I'll check it out soon

nbats:
looks like it only has 11 total playable games so far, I don't think its really worth adding yet given the risk of posting it too.

DogancanYr:
Ingames can also be played.
They're just not flawless.

nbats:
I just think its needs more time to develop before its really something most people are gonna be looking to do. Even PS3 emulation isn't perfect yet so no need to rush into ps4 so early.

PR #3332, opened 2025-04-24
File-hosts and Video File Hosts fixes

clownmedy:
**Changes/fixes:**

- uguu was separated from Catbox in the "File Tools - **File-Hosts**" section. Why were they together in the first place? They don't have even remotely similar limits.
- Upload limitations of uguu were updated in the "Video Tools - **Video File Hosts**" section. The limit is 3 hours, not 3 days.

**Removals:**
- MirrorAce was removed from "File Tools - **File-Hosts**". They deactivated the uploads ability a long time ago, it has never been reactivated.


nbats:
uguu is what catbox was based off of so that's why they were together, but I'm probably just going to remove it from there anyways, I think its mostly useful as a short video host.

I'll also get the limit updated + mirrorace removed

PR #3340, opened 2025-04-26
added wikiloc.com

LuzEterna:
> Trails of the World
> Wikiloc is a place to discover the best outdoor trails for hiking, cycling and many other activities
> We are 17,109,111 members sharing 63,066,958 outdoor trails and 115,518,408 photos.

As the quote suggests on their site, it is a site that allows anyone to contribute in gpx format, supported by images, and is available on platforms such as android, ios and has a very affordable premium subscription.
I always search for my hikes here first to see if anyone has done it before and there is always someone. I can't believe it's not on fmhy.net when I search.

https://www.wikiloc.com/


nbats:

PR #3343, opened 2025-04-26
Remove CRACKSurl from Software Sites

AhmedDawoud3:
ESET blocks the site: "This web page may contain dangerous content that can provide remote access to an infected device, leak sensitive data from the device or harm the targeted device."

Virustotal: 7/97 security vendors flagged this URL as malicious https://www.virustotal.com/gui/url/4f80df32ca1d0cc4d40f33fa70134743e76ec9ba58762576b601c656db6d9e5b


nbats:
I'm sorry but that doesn't prove anything whatsoever. People falsely report sites all the time. We need actual proof of malware, not just a claim.

nbats:
Keep in mind too virustotal is useful more so for scanning actual individual files rather than sites. Scanning a site can give lots of false positives easily

PR #3344, opened 2025-04-26
Update non-english.md

DogancanYr:
FullHDFilmizlesene used to really stream 4k but now it's just text or high bitrate means.
Dizipal link changes daily, it makes more sense to show it via google.
SeriesAZ offers dubbed in Turkish, but also in English, Azerbaijani and Russian.
Added anime sites themselves make and add Turkish subtitles.
And a little correction


nbats:
I also just noticed we dont even have it listed for steam ourselves, so I see no reason to remove it. It works for other platforms still I think.

PR #3369, opened 2025-04-29
Bangla downloading and Bangladeshi  AI sites add to non-english.md

MrCyberSentinel:
MLWBD movie downloading site.
Shothik AI (Bangla AI).
These two things add to the non-English section.

{873152A2-B7F7-47F4-9212-1720EDF7B97F}


MrCyberSentinel:
If you say to remove the AI section, I will remove it. I think it shouldn't be removed because it's Bangladeshi AI, I think, which should be in the Bangla section. Thanks.

nbats:
thanks for sharing, we'll check them out

PR #3371, opened 2025-04-30
Update videopiracyguide.md

WangJian9522:
Added a movie website: indielens.org


nbats:
thanks for sharing, its on our discord queue now so we'll test it soon

PR #3394, opened 2025-05-04
Adding Yampi in Streaming list

devish-ak:
Streams load very fast. ✔️
Content is 1080p, and can be lowered. ✔️
Finds streams consistently for both new and old content. ✔️
They have 9 sources, and it does auto-switch if one fails. ✔️
Site does have episode auto-next. ✔️
Very nice UI, search works well, lots of user settings. ✔️
You can add, search or adjust subtitles on the site. ✔️
It has movies, shows and anime. ✔️
Site has watch lists + continue watching, and you can login to sync them across devices. 🟡
Site has good sorting + random suggestions and editor picks. 🟡
Site is free from new tab/window popup ads, can be used without an adblocker. ✔️
Site looks good + works well on mobile. ✔️
Site does have a custom player. ✔️

Scored: 24 🤩 [i hope so..]


nbats:
Thank you for sharing / grading I appreciate it. The sites UI does look good, but we will still need to check it out and grade ourselves before adding. 

I'll put it high on our test list and we'll try it when we can.

PR #3396, opened 2025-05-04
Update adblockvpnguide.md

DogancanYr:
👍


nbats:
not sure these are worth adding, the crack isn't maintained + the feature list could be found by people easily

PR #3400, opened 2025-05-05
Added Ventoy to create multi bootable usb drive

MyselfRoshan:


nbats:
Thank you but we decided against adding this in the past as its known for cause issues + we have better options already

PR #3407, opened 2025-05-06
Update non-english.md

montarcannon:
add vider.info and vider.cx services


nbats:
we'll check them out, thanks for sharing

PR #3413, opened 2025-05-08
Add MagicTask

heymichaelgeorgiou:


nbats:
ty we'll check it out soon

PR #3418, opened 2025-05-10
Update non-english.md (replaced fake link)

sami155:
replaced fake link


nbats:
it looks like both are very active, we will check out the new one, but we need a better reason to remove the other.

eyestorefivem:
Your approach to this problem is very elegant. I'm learning a lot from reviewing your code.

PR #3420, opened 2025-05-11
add Song/Album Distributors section with SoundOn link

JPShag:
Soundon is run by bytedance and distributes music for free

Also requesting Discord invite - vcpu


nbats:
ty we'll check it out

PR #3425, opened 2025-05-11
Update gaming-tools.md

DogancanYr:
There are still gaps but more can be found if the community helps.


nbats:
thank you for sharing, I don't really know much about this myself, but I will pass it along to testing and see what people think

PR #3443, opened 2025-05-14
Update android-iosguide.md & linuxguide.md

007zzzz:
aurora store is superior

the BSD section is on point, Linux =/= BSD

the 'Others' section may need debloating, removing the commercial and irrelevant OSs


nbats:
- As far as aurora goes we need more reason to bump it up, I think most people would rather just use a site.
- Not sure about a BSD section but we will check out Awesome BSD.
- I appreciate the work you put into the OS section but we just let sites like distrowatch and others handle them.

nbats:
Alright I'll see how people feel about putting aurora #1 

we may make a bsd section just not sure yet

PR #3444, opened 2025-05-14
/ Others

007zzzz:
forgor this one last pull


nbats:
I covered the first 3 edits in the last pull, but idk why we'd add "others", it just makes it longer for no good reason

nbats:
I mean we dont generally add custom OS anyways I'm sorry, its just too much stuff at once

nbats:
I mean we'd have to check all of them, its just too many at once and not something we generally add.

As far as bsd placement goes I understand your point but at the end of the day we have to put it somewhere.

PR #3449, opened 2025-05-15
Update readingpiracyguide.md

buzz35:
added new sources


nbats:

PR #3453, opened 2025-05-16
Update ai.md

Lazymindz:
Added a new tool aboutmepro under text generators. It's a free tool for AI-powered bio generators for every platform and purpose


nbats:
ty we'll check it out soon

PR #3473, opened 2025-05-18
Update gaming-tools.md

DogancanYr:
I don't know what R1 or R2 means.

Also, can we get the original link to Goldberg, it's quite old.


nbats:
I'll get both you added moved to testing, not sure what you mean about old goldberg though

nbats:
I dont think thats a good idea, I've never heard anyone say its confusing

nbats:
their page has downloads, info, help guide and more, its their main git, theres no way we'd remove it

PR #3474, opened 2025-05-19
a small fix and few updates

MathewJake365:
Added the Wayback Machine and its tools to Archive.org, since it is a part of the site.

Added the 'Tero Subtitler' Git repo and bumped it down a step since it's still in pre-release.

Changed 'Subtitle Workshop' to 'Subtitle Workshop Classic' to match the name on its SourceForge page.


MathewJake365 (commented):
Actually I can't outright disagree. I wish there were a better way, maybe something like a tree view for these kind of cases.

nbats:
Yeah I agree with The, I think its fine as is, but we will check out the new site

PR #3479, opened 2025-05-19
Add Namekit in Domain Availability category

yuvadm:
Verified this is a new tool, doesn't exist in the lists


nbats:
thanks for sharing, we'll get it tested soon

PR #3510, opened 2025-05-26
Update beginners-guide.md

geekeie:


nbats:
we dont even have it in fmhy, why would we put it here?

PR #3545, opened 2025-06-02
Add pages.gay as a static site host

lucy-faggit:


nbats:
ty it'll be moved to testing in discord

PR #3549, opened 2025-06-03
Update AI benchmark links and descriptions

ons96:
Added some new benchmark sites to AI Benchmarks (mostly for vision/text-to-image models, and moved VBench url from Video Generation section to here); added an AI Coding Benchmarks section (and moved some of the links in Coding AIs leaderboards to here - I only added the sites that are currently up to date). Also removed Big Code Models Leaderboard from AI Coding Benchmarks as it appears not to be very up to date.


nbats:
thanks for sharing, I moved everything to testing in discord so we'll check them out

PR #3558, opened 2025-06-04
Add free charts maker into Data visualization tools

OwenL1205:
I found MakeCharts today and used it, its ui is very clean and simple, no ads, so want to add it into the list.

I created an issue before, please just ignore it.

FMHY is very helpful, thanks for your work to make things easier.


nbats:
added to our discord test channel, ty for sharing

PR #3564, opened 2025-06-05
Add `frdl` timer bypass

USLTD:
Added timer bypass for `frdl.*` (FREEdlink) file host.


nbats:
added to discord for testing, ty for sharing

USLTD:
> added to discord for testing, ty for sharing

Understood. Where can I get feedback regarding it?

nbats:
you can join here, theres a thread in testing now 

https://discord.gg/MSfjW8KAv4

nbats:
actually this was tested so we'll get it added in the next update, sorry it took so long

USLTD:
> actually this was tested so we'll get it added in the next update, sorry it took so long

No problem. It's actually better if you test it as much as possible.

Additionally, I also just pushed new version which only updates domain matcher (to help **GreasyFork**) and replaced `jQuery` with it's slim version.

For some reason it doesn't work without `jQuery` and before I investigate it a bit more deeply, it will be implemented with it (sorry for end users).

PR #3572, opened 2025-06-07
Update android-iosguide.md

marcelisz:
I have added Continuum and Slide Reddit clients to "Social Media Apps" section. These are new Reddit clients for Android that are forks of Infinity for Reddit and Slide for Reddit apps, respectively, which allow user to enter the API key on the first run, without patching the APKs with ReVanced, and have some additional fixes not included in original Infinity and Slide for Reddit.


nbats:
thank you I'll get it adjusted in next update

PR #3576, opened 2025-06-08
Small fix

marcelisz:
Changed `Go Speed` to `Gopeed` in "Download Managers" section.


nbats:
both are official names, go speed is listed on their git, which we feel is much better for a name

PR #3577, opened 2025-06-09
Add StarOps

miriamhere:


nbats:
Thank you but not sure what this is or if it'd fit anywhere, if you wanna explain it further feel free

PR #3592, opened 2025-06-11
Update img-tools.md

magic-ant:
Add new remove watermark tool


nbats:
thank you we'll check it out

PR #3603, opened 2025-06-12
Small Update 

MathewJake365:
1. Moved line 271 from downloadpiracyguide.md to videopiracyguide.md since they're basically movie and TV download sites.

2. Removed DDLBase.net, as it also redirects to the .com domain.

3. Added RapidMoviez (rmz.cr) to the same line—it's a similar site to HDEncode, but more popular.


nbats:
thanks for the suggestions

1. Those are RD only sites, so I think keeping in RD makes more sense
2. Got DDL base .net removed
3. Added rmz w/ the other RD sites
4. Moved Discogs

PR #3636, opened 2025-06-19
feat: added little fox chinese

AyaanAhmed25:


nbats:
thanks but I'm not too sure this fits, we generally don't add sites aimed towards children only

PR #3651, opened 2025-06-21
New link

sbwgg:
Site that should be on the list


nbats:
moved to discord for testing, ty for sharing

PR #3666, opened 2025-06-22
Removed Extensions for Aidoku

jjoshiii:
Screenshot_20250622-181912
Aidoku announced that they will no longer promote extension sources on their Discord and removed the links to all existing ones.


nbats:
I'm a bit mixed, it says they're separating themselves from unofficial sources which is fine, but do we really have reason to ourselves if they're working?

joshi903:
Aidokus official source was hosted under https://raw.githubusercontent.com/Skittyblock/aidoku-community-sources/gh-pages/. The [Github Readme](https://github.com/Skittyblock/aidoku-community-sources) mentions that the source is no longer maintained but there is a continuation maintained by the community https://aidoku-community.github.io/sources/index.min.json

nbats:
alright thanks for the info, I'll get the discord removed and we'll check out the community one

PR #3679, opened 2025-06-25
Remove obsolete tool from "VSCode Tools" section

Botato300:
Removes [github-vscode-icons](https://github.com/dderevjanik/github-vscode-icons), a browser extension that adds vscode icons to github.com.
It has not been maintained for 4 years. Currently, there is an alternative, [material-icons-browser-extension](https://github.com/material-extensions/material-icons-browser-extension) which is already added to FMHY!

### DIFF
--- a/docs/devtools.md
+++ b/docs/devtools.md
@@ -610,7 +610,6 @@
 ## ▷ VSCode Tools
 
 * 🌐 **[Awesome VSC Extensions](https://hl2guide.github.io/Awesome-Visual-Studio-Code-Extensions/)**, [2](https://marketplace.visualstudio.com/) - VSCode Extensions
-* [github-vscode-icons](https://github.com/dderevjanik/github-vscode-icons) - VSCode Icons
 * [chatgpt-vscode](https://github.com/mpociot/chatgpt-vscode) - VSCode ChatGPT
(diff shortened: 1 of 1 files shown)


nbats:
thank you but tbh is this really something that needs updates anyways? Sometimes things stop being maintained just because there isn't much left to do, and I could see that happening with an 1000+ icon pack.

Botato300:
Apart from the fact that it is not maintained, I forgot to mention something very important: the extension no longer works!
image

> is this really something that needs updates anyways?

Of course, this type of software requires regular maintenance. The extension for adding icons analyzes the HTML of github.com, searches for an ID or class name to know where to insert the icon. What happens if one day github.com changes the part of the HTML that you use? The extension breaks and stops working. It's very common, in fact.

Botato300:
> Apart from the fact that it is not maintained, I forgot to mention something very important: the extension no longer works! image
> 
> > is this really something that needs updates anyways?
> 
> Of course, this type of software requires regular maintenance. The extension for adding icons analyzes the HTML of github.com, searches for an ID or class name to know where to insert the icon. What happens if one day github.com changes the part of the HTML that you use? The extension breaks and stops working. It's very common, in fact.

@nbats

PR #3686, opened 2025-06-25
Update magazine section with 2 websites

freemagazinesnet:
Update magazine section with 2 more reliable websites with weekly and monthly magazines which some times can't find in most popular websites.


nbats:
moved both to discord for testing, thanks for sharing

PR #3722, opened 2025-07-02
Add image editor to  img-tools.md

idotj:
Include Krita in the list for Editing Software


nbats:
thank you but we do have krita already added

PR #3724, opened 2025-07-03
Removed some dead links

Sid954:


nbats:
got them all removed, I would of merged this instead just didn't see it in time, sorry about that

PR #3726, opened 2025-07-03
Remove "Jiffy" Because Chrome Flagged the Extension as Malware

DaveLak:
I had the Chrome Extension installed and was notified by Chrome that it was disabled because "it contained malware" (which I did not verify myself fwiw.)

The chrome extension was pulled already: https://chromewebstore.google.com/detail/jiffy-reader/lljedihjnnjjefafchaljkhbpfhfkdic

Screenshot from my Chrome instance that had it:

image


nbats:
I'll get it removed thanks for the heads up

PR #3747, opened 2025-07-06
Update edupiracyguide.md

Foued80:
FILECONV length converter /Science Math /Physics


nbats:
moved to discord for testing, ty for sharing

PR #3779, opened 2025-07-13
[REQUEST] Add UzTracker as starred ( read description )

helix-nuked:
Add UzTracker as starred since there is m0nkrus uploads and I'm from Uzbekistan and can confirm that the site is legit


ell0uz:
plus asilmedia requires vpn

nbats:
moved to our discord for testing, thanks for sharing

PR #3797, opened 2025-07-18
Update non-english.md

gogoanimex:
added Anime Slayer Web , two times but still not accepted why ??


nbats:

PR #3811, opened 2025-07-20
Add Embeddium to gamind-tools.md

nesuwu:
I thought it belongs here, since its a modified fork of sodium, forked from last stable FOSS (LGPL-3.0) version


hugo9655:
it was decided a while ago (nov 2024) that it wasnt worth keeping individual optimization mods as they're very easily found on mod indexes, including performance modpacks.

this was in the wiki at the time but was ofc removed in the end.

Issue #3837, opened 2025-07-27
Add Site


N8-Yue:
### Type

Site suggestion

### Add additional context

https://appdoze.net/

It gets listed as a trusted site in the r/Piracy Megathread

RottenCarrot21:
I made a pull request

nbats:
thank you but they've had issue in the past with things like unsafe versions of star dock 11 so we decided against it

PR #3838, opened 2025-07-27
Update downloadpiracyguide.md

RottenCarrot21:
Add website mentioned in #3837


nbats:
thank you but they've had issue in the past with things like unsafe versions of star dock 11 so we decided against it

RottenCarrot21:
ok

PR #3868, opened 2025-08-03
Added PixiEditor to Image Tools/Editing Software

PavelDobCZ23:
This is an interesting open source project that caught my eye, because it features a node based procedural editing. It aims to be capable of general 2D editing with raster and vector, static and animated graphics in mind. The most recent update 2.0 has added a whole bunch of features that make it just that. More info can be found on their blog posts.

I would add it to the list of `Image Tools/Editing Software`, because it seems to be quite a generic editor, not focused on one particular use case (like Krita with painting), but rather to serve all sorts of graphical editing needs.


hugo9655:
it looks like this was rejected previously, people were running into a lot of issues at the time.

but if its working well for u now then we can get it re-tested.

PavelDobCZ23:
I would recommend retesting it. Idk when it was first submitted here, but I've only found this app recently, it's got some press after its latest update. So I gave it a try and it seems like quite an interesting project and so far I'm yet to find any noticeable issues.

nbats:
fair enough we'll take another look, thanks for the suggestion

PR #3873, opened 2025-08-05
Improvings and updates in italian section

PeppeMonster:


PeppeMonster:
@nbats can i ask why did you removed proton an bitwarden authenticator?

nbats:
things like that can be found in the normal sections, we dont need to overlap things in every non-eng section

PeppeMonster:
i know, this is why i committed on adblockvpnguide.md and not on non-english.md

nbats:
proton one we're waiting til its further developed, and I moved the bitwarden one to discord for testing

PR #3882, opened 2025-08-06
Adding Singify/FineVoice 

fineshare:
Add Singify and FineVoice to the file


nbats:
thanks for sharing, both have been moved to discord for testing

PR #3902, opened 2025-08-09
Update search.md

gre4test:
Fixed error where there were only 2 dashes above the Search Page Backups section, making the two dashes show and not look like a horizontal rule.


gre4test:
Duplicate of #3743

PR #3914, opened 2025-08-10
tag fix and dead links

Icyvert:
changed all signup tags to "Sign-Up Required" for consistent tags and removed some dead links
also replaced cinemaos home page links to download page ones in the download section for video piracy


nbats:
has conflicts so I can't merge

PR #3915, opened 2025-08-11
Improvings in italian section

PeppeMonster:
added notes about mediaset infinity
missing Live tv on raiplay
compacted anime sites


nbats:
looks good but no need to compact anime sites, that implies they're somehow related to one another but they're not

PeppeMonster:
i see, i thought that since they had the same description they could be merged on the same line.
i'll undo

PR #3918, opened 2025-08-11
Update android-iosguide.md

jamiehunt:
Adds Mindful Slumber to iOS Relaxation tools.


nbats:
thanks for sharing, passed it onto discord for testing

PR #3921, opened 2025-08-12
Update adblockvpnguide.md

usernamex:
Add Proton Authenticator.


swastiksagar:
add https://medeberiya.site/

nbats:
thanks for sharing but we decided against proton for now until its further developed, and we'll check out medeberiya too

PR #3931, opened 2025-08-14
Add Mathos AI

maywanghere:


nbats:
thanks for sharing but last we tried it was pretty limited, like 10 month or something

PR #3938, opened 2025-08-16
Remove TV flag from Einschalten

webstreamr:
That source only has movies, no tv shows.


nbats:
will do ty for the heads up

PR #3942, opened 2025-08-17
Remove KinoKiste

webstreamr:
Redirects to streamtaxi (the next entry) already for a while now


nbats:
got it removed, thanks for letting us know

PR #3950, opened 2025-08-18
Added new SlideShare download resource

bloggerhub148:
Included an additional SlideShare tool [DownloadSlides](https://downloderslides.com/) in the resources list.  
This provides users with another option for downloading presentations in PDF/PPT formats.  
The goal is to give readers more variety and accessibility, similar to the other SlideShare tools already listed.


nbats:
thanks for sharing, we'll send it to discord and check it out

PR #3959, opened 2025-08-20
Update non-english.md

carryingwater:
- [Brazilian Private Trackers Index](https://pirataria.link/trackers#%F0%9F%94%B0-%E2%9E%9C-trackers-brasileiros)


nbats:
like I said its in our discord we'll check it when we can but ty for sharing

PR #3979, opened 2025-08-26
Update file-tools.md

freextool-com:
Online PDF Toolkits Added


nbats:
we'll check it out thanks again for sharing

PR #3980, opened 2025-08-26
Update text-tools.md

bnbroxi:
Emojis Tools Added


nbats:
you dont need to paste each individual tool we'll check out the main site when we can

PR #3986, opened 2025-08-28
Update ai.md

duanbg98:
Add ttsfree.com to the list, it is also a great free tts site.


eli32-vlc:
Sorry this has been rejected / removed before.

eli32-vlc:
I don't the control but its most likely going to be rejected.

duanbg98:
I found it much better than the other sites listed below. It offers up to 500,000 free characters per month. Just simply paste the text and convert.

PR #4003, opened 2025-09-02
Added a very nice expense sharing tool

zizouet:
I added a free expense sharing tool available on Android and Apple Store, it is a very nice one.


nbats:

PR #4009, opened 2025-09-04
Update gaming.md

d3ks1:
Replaced a broken site with a working alternative that was still missing from the list


nbats:
thanks for the heads up, we'll get the dead removed, and check out the new when we can

PR #4029, opened 2025-09-08
Add MyNextInk

david-lo-dico:


nbats:
moved to discord for testing, thanks for sharing with us

PR #4032, opened 2025-09-08
Update file-tools.md

Starship3893:
pixel drain keeps logs of users and share them in police requests.

check https://pixeldrain.com/abuse "police requests"

Keeping logs is a bad privacy policy.


nbats:
lots of file sites comply with police request for illegal porn, its not uncommon. Buzzheavier almost got shut down because they tried to ignore similar things, but they aren't like DMing police officers the IPs of pirates.

PR #4038, opened 2025-09-09
Add Nativefier to Webpages > Desktop Apps

AnonymousElixir:
I've added nativefier as a webpage to app tool


nbats:
thanks for sharing but this got removed because its archived on on github.

catsizedcoder:
I see, Thanks for the swift reply! @nbats

PR #4040, opened 2025-09-10
Add CompliMate AI to AI tools list of ai.md

Complimate-ai:
This adds CompliMate AI, an AI compliance suite for privacy policies and global regulations, to the AI tools list.


FestiveByteRider:
According to their own github repo this tool isn't free and is actually $49/m per domain. Is also their own tool which feels slightly biased to me. ¯\_(ツ)_/¯

nbats:
yeah sorry but if its not free it doesn't really fit FMHY

PR #4050, opened 2025-09-13
Update gaming.md

7shiki:
add retro games nexus


nbats:

PR #4052, opened 2025-09-13
Add SkyViewer resource for sky exploration

111100001:
Added a new resource for interactive sky exploration.


nbats:
thanks for sharing, moved to discord for testing

PR #4061, opened 2025-09-15
Update mobile.md

GuiltiaSinJ:
Added and Starred Discoverium, Discoverium is a fork of Obtainium. Obtainium is great, but it is missing one critical feature. That feature is a search function for discovery of new apps, Discoverium then works the same as Obtainium but with the added benefit of having a search function making it work like a app store but you can get app updates straight from the source.


nbats:
thanks for sharing, we had a test thread in discord but no one seemed to feel strongly, I'll re open it with what you said here

GuiltiaSinJ:
Okay, thanks

PR #4073, opened 2025-09-18
bio.link is no longer free. Removed from internet tools

dadpenguin:
bio.link changed to 7 day free trial.

### DIFF
--- a/docs/internet-tools.md
+++ b/docs/internet-tools.md
@@ -69,7 +69,6 @@
 
 ## ▷ Link in Bio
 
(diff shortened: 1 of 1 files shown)


nbats:
it still says its free on their site, you only have to pay to use custom domain

image

dadpenguin:
image

You are hit with a paywall when you create an account or anywhere when creating a project. I believe the FAQ is outdated. I would recommend to check it out.

nbats:
thank you will do

PR #4130, opened 2025-10-05
Add BiniLossless/Squid link to Audio Streaming

uimaxbai:
Looks like I did something

- Added BiniLossless/squid.wtf counterpart to streaming as it can do that too

Signed-off-by: Max Bai (max@binimum.org)


nbats:
thanks for sharing, we'll check the first, but the squid tidal page is linked on squid homepage.

PR #4150, opened 2025-10-09
Remove U+2060 invisible characters from links

jbgy:
Removed stray U+2060 (Word Joiner) characters from titles and/or links


nbats:
this is conflicting currently, and tbh I just don't know if its really worth keeping up with, its not something 99.9999% of people are gonna notice so its not really worth all the work copy pasting on my end.

nbats:
I appreciate all the effort I'm just being honest

PR #4151, opened 2025-10-09
Update file-tools.md

Far-Se:
Added Regex File Renaming to Data Automation


nbats:
thanks for sharing, moved to discord for testing

PR #4163, opened 2025-10-12
Added Lavender Photos

hastadhana:
add Lavender Photos link under Image Galleries


nbats:
thanks for sharing, we'll move to discord for testing

PR #4164, opened 2025-10-13
Add Tabame as a taskbar alternative

Far-Se:


nbats:
moved to discord to test, thanks for sharing

PR #4166, opened 2025-10-13
remove updateme from android apk tool

phucisstupid:
Updateme is deprecated: https://github.com/anfreire/updateMe-Mobile#update-me


nbats:
thanks for the heads up, will get it removed soon

PR #4168, opened 2025-10-14
added 4 new pages to the single-page route

ShunquanWang:
Supplement the missing documentation in the single-page api


hugo9655:
this is not necessary

PR #4182, opened 2025-10-15
Update mobile.md

DogancanYr:
I believe MiXplorer and MT Manager have the same features, and the note is also important.


DogancanYr:
I think you forgot this note.

PR #4190, opened 2025-10-17
Update m0nkrus

dvuzu:
I have fixed the scraper, now with the new monkrus subdomain


nbats:
thank you but its already updated to new

PR #4194, opened 2025-10-18
Update educational.md with info about Olympiads

Blackbird65536:
There's some more resources on Olympiads on line 1343.


nbats:

PR #4204, opened 2025-10-23
Add trainers category

DogancanYr:
They were grouped into the Trainers category.


nbats:
idk if theres enough for a whole section, it seems fine as is

DogancanYr:
There are more in other megathreads, I can add them if you want.

It should also be separate from the mods category because they don't add anything to the game and some people only want to use mods, while others only want to use trainers.

nbats:
fair enough I'll take a closer look when I can and see if there's enough

DogancanYr:
Should I add other sites?
I don't know if they are looked at or not.

nbats:
you can send to discord but we need to test them all

PR #4205, opened 2025-10-23
Receiving real-time notifications in Thunderbird.

DogancanYr:
Receiving real-time notifications in Thunderbird.


nbats:
thanks for sharing, I'll get it added as a note to our official notes page, then add that next to thunderbird

PR #4211, opened 2025-10-25
revert halloween theme for end of october

SamidyFR:
end of october coming up so :P

Note for nbats - accept this whenever you feel fits, if im correct as long as the theme files arent directly edited (which probably wont happen lol, hopefully no one makes a PR for the theme these next few days) then u can merge without any of the docs being reverted


LandWarderer2772:
should just comment out the code for halloween instead of removing

PR #4221, opened 2025-10-27
Remove GeoSpy from image-tools.md

RottenCarrot21:
Remove GeoSpy because it is no longer allowing public use, only for law enforcements etc


nbats:
thanks for the heads up, I'm editing currently, but I will get it removed in next update

PR #4224, opened 2025-10-28
Update Freedium link for Medium paywall bypass

Sovan22:
I added the working freedium link, got it from the codeberg repository of freedium
https://codeberg.org/Freedium-cfd/web/issues/54

Also, wanted to ask where is the right place to add this app
https://github.com/Sovan22/Medium2Freedium
This is an android app I made to make opening medium links in android a bit easier by just sharing the link to this app and its using freedium to show the article. Just removes the hassle of copying and then opening the link in a seperate browser.

PS : Let me know if it can be added ❤️


nbats:
thanks for the update, got it merged + we'll test the other you sent when we can

PR #4240, opened 2025-11-01
Aseprite: my flatpak build manifest repo

krakua0:
You can review the flatpak build manifest yaml [here](https://github.com/krakua0/aseprite-flatpak/).

I avoided including any of the non-freely licensed materials, so most of the code and art is just pulled from the Aseprite sources.


nbats:

PR #4242, opened 2025-11-01
Update DDLBase description to include 3D and Remuxes

Vojtak42:


nbats:
will add in next update, thank you

PR #4247, opened 2025-11-01
Update gaming-tools.md

Alonso287:
Removed the Minecraft Wiki description text because it was redundant and added [MinecraftTAS](https://minecrafttas.com/)


nbats:
thank you, we'll get the redundant desc removed + the thing you sent tested in discord

PR #4248, opened 2025-11-01
Fix broken discord link for lucida

feenko:


nbats:
thank you we'll get it fixed in next update

PR #4269, opened 2025-11-06
Update OneHack with new URL

catsizedcoder:
I have added the new domain as a secondary url as their site has a popup when visiting that says "To be safe, if our site ever goes offline, please remember our new domain:
onehack.se" as shown below.
image


nbats:
I  mean it just redirects to .us right now, but we'll replace it if it ever changes

PR #4276, opened 2025-11-07
Add pdfarranger to PDF tools list

Jaysum57:
pdf arranger is a Small python-gtk application, which helps the user to merge or split PDF documents and rotate, crop and rearrange their pages using an interactive and intuitive graphical interface.


nbats:
thank you but its already in fmhy

PR #4279, opened 2025-11-08
Update packages and add docs for API deployment

Pasithea0:
Updated all the packages to their latest compatible versions and add documentation for selfhosting the API. wispy and I learned many things lol


nbats:
Thank you much appreciated

PR #4280, opened 2025-11-08
added a reasonable nginx config as example in the selfhosted page

rhld16:


nbats:
Thank you got it merged

PR #4287, opened 2025-11-11
Add WizFile to file tools list

PFCPRO:
I feel if WizTree is included, WizFile should be as well


nbats:
thank you but it is listed already under file tools > disk tools

PR #4291, opened 2025-11-11
Add Dizipal and Dizilla to non-English resources

DogancanYr:
The sites remain active. As Dizipal’s URL frequently changes and the number of links continues to grow, access is being redirected via Google.


nbats:
not sure on the google results, but we'll add the site itself

DogancanYr:
Ok.

PR #4292, opened 2025-11-11
ToS;DR with Extensions information

DogancanYr:
Updated ToS;DR to include Extensions for better resource access.


nbats:
they have this link at the top of their site, I dont think we need it ourselves

PR #4315, opened 2025-11-18
Adjust names for some tools as well as remove incorrect discords.

bradleytechman:
- The checkra1n discord is incorrect, that discord is for general iOS hacking, not checkra1n support. 
- The TrollStore discord is wrong, I am a moderator for the subreddit and we don't have a discord, the one that was linked was the discord server for a separate app (that app is linked in the TrollStore line, however it's misleading to link that discord)
- the iOS guide is already starred lower in the guide, and also rename it because it's not a CFW guide, it's a general jailbreaking guide.


nbats:
Thanks for the updates, just got it merged

bradleytechman:
np

PR #4332, opened 2025-11-20
Add Proton Sarek to Linux gaming resources

mostafatouny:
Gamers with old GPUs cannot use standard Proton or Wine, and are advised to try Proton Sarek.


nbats:

PR #4334, opened 2025-11-20
de-star rarely used tools

mostafatouny:
The following won't be used by at least 95% of linux gaming users, and thereby their star is removed.
* proton-ge-custom
* ProtonPlus or ProtonUp-QT
* Kapital Sin
* Torrminatorr


nbats:
That's fair but we'll need more detail to unstar these, the best way is to come to our discord and make star removal threads

https://github.com/fmhy/FMHY/wiki/FMHY-Discord

PR #4350, opened 2025-11-26
Add a best self purity guide game

Taylojohns:
The best version of Rice Purity Test.


nbats:
Thanks for sharing but I don't see what it has to do with WoW

PR #4369, opened 2025-12-02
Add CloudConvert to file conversion tools list

LordAizen1:
### Changes
- Added **CloudConvert** to `docs/file-tools.md` under the "File Converters" section.

### Context
CloudConvert is a popular and versatile online file converter that supports a wide range of formats (audio, video, document, etc.). It was missing from the wiki and is a valuable addition for users needing quick file conversions.


nbats:
thanks for sharing but isn't 25 daily a bit limited? I it was removed in the past for that

PR #4370, opened 2025-12-02
Update MangaBaka domain from .DEV to .ORG

jippi:


nbats:
thank you its updated now

PR #4383, opened 2025-12-05
Add NepoFlix 

Sandipeyy:


nbats:
correct me if I'm wrong, but this seems to be a frontend for a single API, it does have a nice ui, I just don't know if it stands out source wise.

Issue #4395, opened 2025-12-06
Light theme after December update has bad contrast and non-links can barely be read


Pentaphon:
### Type

Grammar / Markdown

### Add additional context

Light theme after the Winter update has bad contrast and now the entire site can barely be read. See screenshot. 

<img width="731" height="632" alt="Image" src="https://github.com/user-attachments/assets/e0f0c5d2-1878-4e6d-aa48-ac9f293198a0" />

ClaraIsCute:
@Pentaphon Your issue has been fixed. You can close this issue now.

<img width="1199" height="594" alt="Image" src="https://github.com/user-attachments/assets/d8ffa83e-55be-4414-a11f-358d0ac9e3e7" />

PR #4396, opened 2025-12-07
fixed light mode font issue

sirbread:
#4395


sirbread:
whoopsies amp added whitespaces to the file 
kids never do git add .
reverting one sec

PR #4401, opened 2025-12-07
fix equicord and youtube downloaders

Kovionata:


Kovionata:
but I forgot to delete `themeHandler.ts` in my fork, anyway I'm fixed. And I was miss spell at my commits.

nbats:
thanks for the heads up we'll fix it soon

PR #4416, opened 2025-12-09
Add Tachimanga to Starred section

darwin-256:
Tachimanga offers a superior user experience compared to other iOS manga reader apps and supports Tachiyomi extensions, which are the most popular extension sources on Android.


nbats:

PR #4432, opened 2025-12-12
Add davincibox and edit format of DaVinci Resolve

litekin:


nbats:
we try to keep linux tools in linux, but I will fix the formatting like you have it here, and we will test this in discord

PR #4435, opened 2025-12-12
Update ChatGPT version to GPT-5.2 in ai.md

nqlp:
don't forget to update the wiki too:
https://github.com/fmhy/FMHY/wiki/FMHY%E2%80%90Notes.md#chatgpt-limits


nbats:
we switched it back, apparently 5.2 is for the workplace-only version, not consumer

PR #4442, opened 2025-12-15
[Roblox] Fishstrap alongside Bloxstrap

Margeshy:
Fishstrap is a fork of Bloxstrap. It is updated more frequently than Bloxstrap, and has more commits (attow: 1,682 vs 1,293)
It has more features, and changes to existing ones that return fuller information
Fishstrap is under the @fishstrap group, that is majorly maintained by @returnrqt, with some minor contributions by others, compared to Bloxstrap which has 26 listed contributors
Bloxstrap is not under active development currently


WitheredFOT:
I second this.

Overimagine1:
Added back in ad33380

It was initially removed because it was suspected to bypass the FFlag limits Roblox imposed which was leading to bans, similar to what Voidstrap is doing. We realize now that isn't the case as Fishstrap bans people from their Discord for even talking about it.

You or @nbats are good to close this.

PR #4461, opened 2025-12-20
Update audiobook downloaders

aviiciii:
- Removed deprecated downloaders for tokybook.
- Added downloaders for multiple audiobook streamers.


nbats:
thank you but we'll need to test the downloader, we'll move to discord to try it, also the first tokyo book one is updated still

PR #4464, opened 2025-12-20
Docs/fix mobile adblock info

devsiddhant:
Corrected mobile adblock info


nbats:
thank you I'll pass this all along and see how people feel

PR #4481, opened 2025-12-22
Move Vesktop to Vencord line to be consistent with Equicord/Equibop

litekin:
Equibop is the desktop client of Equicord. Vesktop is the desktop client of Vencord.
Equibop is on the same line of Equicord, so Vesktop should be on the same line as Vencord.

### DIFF
--- a/docs/social-media-tools.md
+++ b/docs/social-media-tools.md
@@ -64,9 +64,9 @@
 ## ▷ Discord Clients
 
 * 🌐 **[Discord3rdParties](https://github.com/Discord-Client-Encyclopedia-Management/Discord3rdparties)** - Discord Clients Index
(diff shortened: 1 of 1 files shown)


litekin:
> put Vencord's Github link and Discord server before Vesktop's

I just looked at https://fmhy.net/privacy#privacy and see that lines with multiple links that have Githubs and Discords have the Github and Discord links directly after each link. If that precedent is the "correct" way to do it, than the Equicord line was wrong originally. Here's the corrected commit: https://github.com/fmhy/edit/pull/4481/commits/9644a5c736d7c82acd86a23d1a0522f3c24d6331

TheUltraParanoid (approved):

PR #4485, opened 2025-12-23
Add Aurora Store to the beginners guide

ATG13:
Aurora store is an easier and safer way to install and maintain android apps.


nbats:
thanks for the suggestion, I brought it up in our discord so we'll see how people feel, if they're on board I'll make the edit

PR #4498, opened 2025-12-27
Added MouseCTRL - a free mousepad finder tool

JW-TechTest:
Added MouseCTRL - a free mousepad finder tool and searchable database containing information about a wide range of gaming mousepads


nbats:
thanks for sharing, we got it tested and added earlier

PR #4501, opened 2025-12-28
Updated Eaglercraft Server List Website

RealRigBot:
The original site has a banner saying it's permanently shutting down and will go offline December 31st. TopEaglerServers.com is the next biggest and has hundreds of more active and different servers.

Original site: https://servers.eaglercraft.com/
New site: https://topeaglerservers.com/


nbats:
thanks for the info, we'll get the old removed and move the new to discord for testing

PR #4502, opened 2025-12-28
Update educational.md

awawg482:
CGPA is commonly used in international grading systems and differs from GPA in calculation and scope.


nbats:

PR #4503, opened 2025-12-28
Fix mobile adblock info

ghost:
Fixed the implication that brave also supported uBO on android


nbats:
thanks for the heads up, we'll get it adjusted

PR #4549, opened 2026-01-05
reorder note taking

VVispy:
replaced notion star with appflowy. Notion generates insanely large cache files on windows (about 1gb daily) and not open source. Appflowy can do everything notion can afaik and open source.


litekin:
good pr. it's funny how if you look at Notion's site, it seems they've completely given up on targeting consumers and are now fully corporate slop. there's no reason to use it.

PR #4555, opened 2026-01-06
Update system-tools.md

ankurajha:
Added Printer test online tool


nbats:
no need to send twice we'll test when we can

PR #4575, opened 2026-01-10
Star TuneMyMusic and move to Playlist Tools

litekin:
to be honest I haven't used TuneMyMusic but it appears to be the best option for playlist transferring between services by far, and it does belong in Playlist Tools. I'd open a star suggestion but I have nothing to say about it other than "looks good"

### DIFF
--- a/docs/audio.md
+++ b/docs/audio.md
@@ -331,6 +331,7 @@
 
 ## ▷ Playlist Tools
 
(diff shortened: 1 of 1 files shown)


SamidyFR:
playlist tools is a spotify specific subsection lol.


but about tunemymusic, ive used it myself and i think its 100% worth to star, very high quality site

litekin:
shoot yea you're right :facepalm: in that case I'll close the PR and you make a star suggestion if you don't mind, since you've used it.

PR #4582, opened 2026-01-12
Add Openlib and OpenlibExtended to books section

warreth:
Added links to Unofficial Mobile app and OpenlibExtended.


nbats:
thanks for sharing, we'll get both tested when we can

PR #4587, opened 2026-01-13
Update non-english.md

Protonval:
Movix changed their domain to movix.blog in the french part


nbats:
thank you I got it updated in editor, will show in next update

PR #4606, opened 2026-01-19
Refactor rate limiting logic and improve error handling

mispaints:
Modernize rate limiting by using Cloudflare’s canonical IP source, adding safe fallback, and improving failure handling while keeping the implementation minimal and secure.


nbats:
Don't take this the wrong way because we appreciate the effort but one of our devs said its cf, we don't have a backend, so its not needed.

PR #4610, opened 2026-01-21
new notes site

VVispy:


eli32-vlc:
I do not approve. A dedicated note site can be used as backup. The main problem is this note site can be blocked and its easy cause someone might block pages.dev due to scams etc since its so easy I seen ms defender scams on pages.dev. And if someone misses a note it could be dangerous.

nbats:
we'll add git links back too so people have options

PR #4611, opened 2026-01-21
tooltips

bbbreaddd:
I implemented the tooltips feature from wotaku.wiki (ex. https://wotaku.wiki/websites)
I also made it match the theme color and made it compatible with other sites.
Here's a live example: https://bbbreaddd.github.io/edit/privacy#antivirus-anti-malware
(Look at the end of the bleepingcomputer line)

Here's how you add a note.
Go to the `docs/.vitepress/notes/` folder and make a new .md file with the filename being the title of the note. Then just paste any text in there and copy the github link (ex. `https://github.com/fmhy/edit/blob/main/docs/.vitepress/notes/test.md`). Then you can paste this anywhere in the wiki using a hyperlink and it'll automatically be converted into a tooltip while also maintaining compatibility with other sites. If the link is broken for whatever reason, it'll just convert back into a regular hyperlink.

I updated one note in the privacy.md as an example. I don't have time right now to do the rest. There are still some improvements I want to make, but I believe it is ready to be implemented in FMHY.

Note: I used AI a lot. I did force myself to learn more how vitepress, html, and css and all that stuff works but I'm still a beginner and relied on AI for most of the complicated stuff. 
(I just noticed that one of the files I copied was copyrighted. Don't kill me tasky)

### DIFF
--- a/docs/.vitepress/theme/components/Tooltip.vue
+++ b/docs/.vitepress/theme/components/Tooltip.vue
@@ -1,8 +1,13 @@
 <script setup lang="ts">
-withDefaults(
+import { withBase } from 'vitepress'
+import { computed } from 'vue'
+
+const props = withDefaults(
   defineProps<{ title?: string; icon?: string }>(),
   { icon: '/note.svg' }
 )
+
+const resolvedIcon = computed(() => withBase(props.icon))
 </script>
 
 <template>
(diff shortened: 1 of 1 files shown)


SupremeMuhit:
@bbbreaddd bro give preview screenshot, let us see how it looks

bbbreaddd:
> @bbbreaddd bro give preview screenshot, let us see how it looks

I sent a link where you can test it but here you go

Screenshot_20260121_080451_Firefox

bbbreaddd:
Okay, I think I fixed the copyright problem. I used the floating-vue documentation to make the tooltip, and I also made my own icon for the button. There isn't any code from tasky anymore, so it should be fine to merge. I also added a button that takes you to the original note URL. I only added a few more notes, and I created the individual note files to link to, someone just needs to replace the link for all the notes in the wiki.
Here's how the new tooltip looks like
image
image

I'm still not sure if that's the icon I want to keep, but I don't know much about design so any suggestions would be nice. Also hopefully I did the commits right, still not used to git.

PR #4634, opened 2026-01-24
update multi model section

KHROTU:
update the models to their latest offerings, remove the explicit "multiple models" tag (except the first two, those have way too many models) since this is the "multiple model sites" section afterall


nbats:
this has conflict now they'll need to be resolved

khrotu:
done.
one thing to note is that currently the pollinations website has a broken ui, and even without this ui bug they dont write the full name of models, so the one written are purely based on what was already there and their api

PR #4640, opened 2026-01-25
fix exact search

bbbreaddd:
Before:
image

Now:
image

Preview: https://af4a49c8.edit-b69.pages.dev/

Also added better handling of dashes and underscores when using fuzzy search.

I would like people to test this though to see if it works well. I've done some testing and it seems to work well, but hard to test everything.


bbbreaddd:
Found out that invisible characters that are for some reason inside many places in the wiki can break the search so I added a commit to ignore those.

bbbreaddd:
Added a way to navigate the snippets in the search
https://files.catbox.moe/q8bvih.mp4

bbbreaddd:
Fuzzy search separates each word that matches when highlighted. This makes it hard to navigate using the new feature. To improve this, I made it so nearby highlighted words would merge together. There are also new keyboard shortcuts that allow you to easily go from the next and previous highlight.

PR #4646, opened 2026-01-27
add tooltips to color picker

eli32-vlc:


nbats:
built into browsers already but thank you anyways

PR #4649, opened 2026-01-28
edit jdownloader.md

litekin:


litekin:
idk how this ended up as a separate pr but I think I get it now

PR #4662, opened 2026-01-30
Note edits 2

litekin:
more note edits. I'm not sure how the code block that I added to [the eruda note](https://github.com/litekin/edit/commit/5b7f03553382df2c76f1fd54ec82f4717d87f82f) will look on the site.
also I retract the statement that I would've gotten through all of these yesterday.


litekin:
why is this not showing up in PRs

PR #4664, opened 2026-01-30
Update social-media-tools.md

US-USA:


nbats:
thanks but we're a bit confused on use-case, it says it shows stories, but all profiles just seem to be user info, is that very useful when other frontends / viewers exist like this: 
image

PR #4668, opened 2026-01-31
feat: remove Kizzy from mobile list

rushiiMachine:
Kizzy is a Discord selfbot that does not advertise itself as such, and does not even emulate real client properties (dead8309/Kizzy#56) as is necessary to ensure that Discord accounts are not flagged and banned. The developers have consistently attempted to hide this fact for 4 years by now, and have removed any disclaimers that used to exist about their software being potentially dangerous.


nbats:
ty for the heads up, I have editor open rn so it'll  be removed soon

PR #4669, opened 2026-01-31
Fix missing link on "Android Sticker Packs for Signal"

DEAEq:


nbats:
that was just a mistake we only need it linked once but I'll move it to end of line so it works for molly too

PR #4672, opened 2026-01-31
Fix Animelok link in non-english.md

VivekKavala:
Updated Animelok entry with a new link.


nbats:
ty its fixed both places we had it now

PR #4681, opened 2026-02-01
fix various typos

KHROTU:


nbats:
thank you, I think lo-fi is technically right if I'm not mistaken, but I will add the other changes.

nbats:
nvm I see you switched from low to lo I will do that too

PR #4686, opened 2026-02-01
add tekmods

carryingwater:
- tekmods is a popular brazilian page for mod apks.


nbats:
moved it to our discord for testing, thanks for sharing with us, I'll fix pirataria too

PR #4706, opened 2026-02-04
Update roblox tools

Margeshy:
The section is already pretty incomplete so no stars from me, just some simple adds, rewrites, and reorders

**Added**
[BetterBLOX](https://betterroblox.com/) as an extension, after RoValra (Last Online Status, Server Region Selector, Smart Notifications (Friend activity, game join))
[Sixteensrc](https://apidocs.sixteensrc.zip/) has some great api that were just... undocumented for whatever purpose. These were leaked from a data miner, tell me about the viability of this being added

**Reordered**
RoPro to be last, as some features are locked behind a subscription
Rotrends to be first and in bold. Better ui, more depth in graphs

### DIFF
--- a/docs/gaming-tools.md
+++ b/docs/gaming-tools.md
@@ -1008,15 +1008,15 @@
 
 ## ▷ Roblox Tools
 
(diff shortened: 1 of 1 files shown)


nbats:
image

PR #4715, opened 2026-02-06
Add Windsurf note about SWE 1.5 being free until 24.03.2026

Dima-369:


nbats:
that is cool but its only gonna be free for a month, idk if its really worth noting in its desc

Issue #4720, opened 2026-02-07
Add DDoSecrets


ghost:
### Type

Site suggestion

### Add additional context

whistleblower site founded in 2018 for news leaks

https://ddosecrets.org/article/important-notice-ddosecrets-domain-change

nbats:
thanks but we do have it just without all the extra stuff at the end of the url

ghost:
> thanks but we do have it just without all the extra stuff at the end of the url

<img width="925" height="175" alt="Image" src="https://github.com/user-attachments/assets/a97cd6f7-365d-4f38-8715-54a1ca63f8e5" />

litekin:
<img width="937" height="555" alt="Image" src="https://github.com/user-attachments/assets/c3e2a9d9-deec-4d02-b512-04554b5c6f27" />

PR #4726, opened 2026-02-09
Update reading.md

Biplume:
Renames the DDOS thing - the acronym is better known for other meanings, hence confusion like #4720


SupremeMuhit:
i think the current DDOS is good; everybody knows that DDoS is the short form of Distributed Denial of Secrets
And in that issue they searched for "ddose." That's why no result came. I can see results searching DDos

ghost:
image

I don't think everybody knows that to be frank. This collective is not something that I've encountered before or would expect to be common knowledge at a glance.

The also probably searched for 'ddose' given the URL for the organization starts with that, but in any case I feel like the more common use of the acronym is absolutely not in reference to them.

nbats:
fair enough, will do in next update

PR #4727, opened 2026-02-09
Update video-tools.md

Biplume:
Picrew is already linked and starred in image-tools.md and has nothing to do with video animation - it generally creates still images of characters often used as avatars or reference.


nbats:
thanks for the heads up, will fix soon

PR #4743, opened 2026-02-11
Update SportOnTV links to be more accurate

sportontv:


nbats:
ty I'll get this adjusted

PR #4749, opened 2026-02-12
Fix VegaMovies link in video.md

swamoth:
The current Link for VegaMovies that is https://vegamovies.gripe/  is not working at least in southeast asia region and the updated link is working


nbats:
thanks for the heads up, I'll get it fixed

PR #4751, opened 2026-02-13
Base64 Dialog

bbbreaddd:
Adds a dialog explaining how to use Base64 when clicking on a Base64 link
firefox_U4zVryuRyH


bbbreaddd:
After some changes, I think I'm going to leave it like this
image

PR #4755, opened 2026-02-13
Update ai.md

KHROTU:
gemini - not to be confused with ga model (not released yet) minimax - update model


nbats:
thank you I'll get this adjusted

PR #4765, opened 2026-02-15
Tekmods

carryingwater:
- tekmods is a popular brazilian page for modded apks.
- fixed a little ambiguation in the "pirataria" megathread link.


SupremeMuhit:
tekmods was tested and rejected earlier

PR #4769, opened 2026-02-15
Star ollama + top of the list

Margeshy:
I noticed that Jan and LM Studio are both starred as recommended options, but Ollama isn't, and I'm curious about the reasoning behind that

Ollama has more support, documentation, and knowledge compared to the other two. It's also fully open-source under the MIT license, whereas LM Studio is proprietary


nbats:
we'll pass it on to discord and see how people feel

PR #4786, opened 2026-02-17
Revamp File Converters section

znarfm:
Include platforms and if online or offline


nbats:
looks good thanks for doing this

PR #4792, opened 2026-02-18
Added Go Speed to list of torrent clients as well

FaridZelli:
Go Speed is both a download manager as well as a torrent client with popular trackers included by default.

Already included under download managers:
https://github.com/fmhy/edit/blob/main/docs/file-tools.md#-download-managers


nbats:
I mean is it worth using as a torrent client over better options like qbit, just because it can work as a torrent client doesn't mean its should over others.

FaridZelli:
> I mean is it worth using as a torrent client over better options like qbit, just because it can work as a torrent client doesn't mean its should over others.

It's pretty amazing. The app is fairly nice to use and is supported on:

- 🪟 Windows
- 🍎 MacOS
- 🐧 Linux
- 🤖 Android
- 📱 iOS
- 🌐 Web

Additionally:

- Gopeed includes many trackers by default
- Flatpak version doesn't require access to home directory or entire filesystem unlike KTorrent, Fragments, Transmission or qBittorent
- The project is actively maintained

I stopped using qBittorrent Enhanced a long time ago and switched to Gopeed.

Gopeed had its star removed awhile ago [for unclear reasons](https://github.com/fmhy/edit/blob/e095854583244bcd4da19f9dd080f558f22d6f97/docs/posts/mar-2025.md?plain=1#L65). I may open another pull request to add it back once this gets accepted.

nbats:
alright thanks for all the info, I made a suggestion thread in our discord with a screenshot of what you sent so we'll see how people feel

PR #4798, opened 2026-02-19
Update video-tools.md

Vt3RDYF85v:
I've added few sites for Video Tools as these are related to FFmpeg project.


SupremeMuhit:
sent those links to discord testing; we will review and decide if those worth adding

PR #4799, opened 2026-02-19
Update gaming-tools.md

chinnick967:
Adding Metabot.GG as a resource for MOBAs


SupremeMuhit:
@chinnick967 it was rejected earlier; if you feel strong about it, we can check again

PR #4802, opened 2026-02-20
Update ai.md

Vt3RDYF85v:
- Add GPTChatly
- Add Koala.sh

I've found these sites as this would be useful


nbats:
moved both to our discord for testing, thanks for sharing with us

PR #4803, opened 2026-02-20
update discord link for p-stream

Duplicake-fyi:
I'm one of the p-stream devs and our discord got banned, I've updated the link to point to the new server. The invite link never expires and has unlimited users

25k+ members gone...


nbats:
sorry that happened it really sucks, I will get it updated soon thanks for sharing new link

Duplicake-fyi:
> sorry that happened it really sucks, I will get it updated soon thanks for sharing new link

we've moved to fluxer now, so don't update the link with discord: https://fluxer.gg/3kSqBs0D

PR #4818, opened 2026-02-22
Add Emoji Creator links for multiple languages

almeidamarcell:


SupremeMuhit:
there is no need for adding the site in all non-eng section, if the site gets added

PR #4822, opened 2026-02-23
Changed some icons in homepage

SupremeMuhit:
I changed 4 icons on the homepage. if you dont like any (hope not), just close this PR, and if you have problem with specific, i am here to revert

image


nbats:
I made a suggestion thread in discord so we'll see how people feel, I have no issues personally

PR #4823, opened 2026-02-23
Update gaming.md - Add shadPS4 (PS4 emulator)

gonlad-x:
Add PS4 emulator, shadPS4, to Emulators list.


nbats:
thank you but we dont add emulators ourselves since gametechwiki does such a good job already

PR #4826, opened 2026-02-23
Add Nodal.gg for game discovery and recommendations

Nicholas-Amirsoleimani:


nbats:
ty it went through testing and has made it in, it'll show up next update

PR #4829, opened 2026-02-24
Add new AI source - Lumo

rahmidaniel:
Lumo is a chatbot by Proton. It's privacy focused and good for daily tasks.


SupremeMuhit:
it was rejected earlier
image

PR #4834, opened 2026-02-25
Add Ableton Live to audio.md

Alexuspo:
Added Ableton live to Audio editors and also included a cracker


SupremeMuhit:
sent those links to discord for testing, we will review and decide if those worth adding, thanks for sharing.

PR #4839, opened 2026-02-25
Remove Trackly.cc from torrenting documentation

Union-Crax:
Removed Trackly.cc from the list of torrent sites (Site has been dead for a while, development has stopped).  Side note: development might start again later on, but owner isnt sure.


nbats:
ty will merge with current updates soon

PR #4841, opened 2026-02-26
fix: correct spacing in Gemini CLI description 

nqlp:


nbats:
this added a space to video, but I will do the others, ty

Hesseytrs:
the spacing was correct on the gemini cli, it was meant to show it can use Gemini 3 Flash or Pro, those aren't 2 separate models you can select from what I know, it just chooses which one would be better for the user's prompt hence the "Gemini 3 Flash/Pro" as both are used, the user just can't control which one is used though.
ig it can be just "Gemini 3" but that's vague so putting it back as I originally had it would be best.
Screenshot_20260226_040300_Fennec

nbats:
gotcha I'll keep that the same

PR #4844, opened 2026-02-26
fix convert to it name

eggbagel:


SupremeMuhit:
you are in discord you could just tell the edit in #to-do 😓
i think that would be more easy for nbats (i guess?)

eggbagel:
> you are in discord you could just tell the edit in #to-do 😓 i think that would be more easy for nbats (i guess?)

i suggested it in the star suggestion post but it still is called Convert.to so i thought might aswell make a pull request

nbats:
thank you, I have editor open so I'll combine this edit into that

PR #4858, opened 2026-03-01
Starred TinyWiiBackupManager

Checco763:
All i did was star TinyWiiBackupManager, i think it would be a great option because TinyWiiBackupManager is the only backup manager in the section, and mostly what catches someone's eyes first, are the starred sites


SupremeMuhit:
sent the suggestion in discord for discussion

Checco763:
Oh thanks, I didn't know you had to use discord for star suggestions

SupremeMuhit:
> Oh thanks, I didn't know you had to use discord for star suggestions

yeah every link and suggestion are discussed there then we decide if it worth adding

PR #4865, opened 2026-03-02
Update some iOS jailbreak info and change jailbreak chart URL

bradleytechman:
The old jailbreak chart is no longer maintained and is now maintained by me (it's the same one in r/jailbreak now also, dw I'm not some random person)

There was also just some outdated info and missing jailbreaks so I added those. In the future, it would be nice to maybe specify device support and show the recommended jailbreak but for now it's fine.


nbats:
there seems to be some random text at top + a random space at the bottom

bradleytechman:
that's really weird, it's only there in the PR file but when I go to the actual repo it isn't there. 

The reason it is there is because when I was updating it, I knew some jailbreaks were missing and I had a list of them, and I didn't really want to go through it manually so I just gave chatgpt that section and the list and told it to add any that were missing and somehow that text from chatgpt got caught in there. really weird.

bradleytechman:
anyways, it's fixed now, I've always starred plumeimpactor since another tool by them (feather) is already starred and it's a good tool

flowerey:
(https://verygenericname/nathanlr)

um typo in the link?

bradleytechman:
ok that is really weird cause i specifically typed the link correctly myself with github.com

PR #4870, opened 2026-03-02
added chrome extension and video downloader

korchix:


SupremeMuhit:
sent those links to discord for testing, we will review and decide if those worth adding, thanks for sharing.

PR #4878, opened 2026-03-04
Updated french Torrenting websites

enzomtpYT:
YGGTorrent (the biggest french torrenting website) got hacked and closed
Sharewood is about to close in the at the end of month

UTOPEER scraped all YGGTorrent data before the hack and offers all of yggtorrent without sign-up for now. UTOPEER is a federated project of torrenting website, the domain will close at some point and users will have to create nodes to access torrent later on.

C411 was made for replacement of YGGTorrent but does not have as much torrent as YGGTorrent
Others are private torrent website they don't have as big of a database as YGG had

### DIFF
--- a/docs/non-english.md
+++ b/docs/non-english.md
@@ -463,7 +463,11 @@
 
 ## ▷ Torrenting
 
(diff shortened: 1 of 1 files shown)


nbats:
thank you, it is removed in the editor, I just haven't sent it for today yet but I will soon

PR #4884, opened 2026-03-05
Add C411 to french torrenting

enzomtpYT:
Knowing YGG just closed, sharewood about to close plus they closed registration 5 days ago
Torrent9 only has low quality torrent and lies about peers i suggest to add C411 to at least list an option that will last and good quality


SupremeMuhit:

Issue #4883, opened 2026-03-05
Add closed-source note to YouTube Plus


KipJM:
### Type

Bad sites

### Add additional context

YouTube Plus, aka YTLite (https://github.com/dayanch96/YTLite) is fully closed sourced since 4.0, as stated by the author here: https://github.com/dayanch96/YTLite/issues/244

Since this is not mentioned on the README, a note or warning should be added on fmhy.
https://fmhy.net/mobile#ios-youtube-apps

nbats:
will do, thanks for the heads up

PR #4889, opened 2026-03-05
fix: note YTLite closed source since v4.0

cedev-1:
Added note that YouTube Plus (YTLite) is closed source since v4.0

issue - #4883


nbats:
thank you will do next update

PR #4903, opened 2026-03-08
add Macrium Reflect to backup tools list

0xNikilite:


nbats:
thank you, we'll test it + decide then

PR #4906, opened 2026-03-09
claude chat changes in ai section

KHROTU:
- move claude chat to multiple model sites. its not official and has better models.
- added note on how to use claude chat.
- changed the description of ish from "multiple generators" to "multiple models"
- organised providers based on their rate limit and model available


nbats:
this has conflicts now that need to be resolved before we can consider

khrotu:
@nbats go to https://github.com/fmhy/edit/pull/4934

PR #4911, opened 2026-03-10
remove suspended server

ifrahimm:


nbats:
will remove next update, ty for the heads up

PR #4912, opened 2026-03-10
Revise Buzzheavier expiry information

Union-Crax:
Updated Buzzheavier file expiry details from '15 Days After Last Download' to '2 days expiry + 1 hour per download'. They changed their file expiry system.


nbats:
thank you, I'll merge this into editor soon

PR #4924, opened 2026-03-12
Organize downloading.md

znarfm:
- Add source code / github links where applicable
- Add platform indicators where applicable
- Reorganize some entries like in Usenet > Downloaders


znarfm:
image

I moved these extensions from sabnzbd to the nzbunity entry since the extension homepage leads to https://github.com/tumblfeed/nzbunity

image

PR #4925, opened 2026-03-12
Change Buzzheavier expiry duration to 7 days

Union-Crax:
Updated Buzzheavier expiry from 2 days to 7 days.


nbats:
ty will fix next update

PR #4926, opened 2026-03-12
Buzzheavier no longer supports torrents

Union-Crax:
Clarified the behavior of the download button for Buzzheavier. (Buzz no longer supports torrents)


nbats:
ty will fix next update

PR #4932, opened 2026-03-13
chore: remove hianime

LumiFae:
HiAnime has been retired it seems, so thought it might be good to remove it from fmhy


SupremeMuhit:
i dont think its permanent, they must be planning to rebrand it, but those links are gone for sure.

PR #4938, opened 2026-03-14
moved macos vms to the correct location

hhoneeyy:


nbats:
as far as I know these are macOS for linux, so linux makes sense

hhoneeyy:
oh that'd make sense

PR #4944, opened 2026-03-15
chore: update Perplexica name

First-Non-Interesting-Username:
Updated the name of perplexica, since it has been changed.


nbats:
thank you will fix soon

PR #4953, opened 2026-03-17
Remove VRPirates from VR Games list

kitiki1222:
Removed VRPirates link from the PCVR Games section.

VRP has shut down following a cease-and-desist from Meta.


nbats:
ty its gone now

PR #4954, opened 2026-03-18
Update Komikku link in Linux-MacOS documentation

anmol210202:
Current link is of android version of komikku, makes no sense in linux/mac os page, so here's the updated flatpak version for komikku....


nbats:
thank you, will merge in next update soon

PR #4955, opened 2026-03-18
Cromite browser doesn't support extensions

ludrol:
source: https://github.com/uazo/cromite/blob/master/docs/FAQ.md#does-cromite-support-extensions-in-android


nbats:
thanks for the heads up, we'll update description soon

PR #4978, opened 2026-03-21
Added highlited after search result clicked. Fixed scrolling bug and added a button to disable / enable the hightlights.

eli32-vlc:


bbbreaddd:
Improved scrolling (and removed highlighted lines because I don't like the way it looks, just my opinion though)
https://github.com/bbbreaddd/edit/tree/search-highlight-fixes

PR #4981, opened 2026-03-22
Updated the lists of process managers

Kolleger:
I removed System Informer from Process Monitors under System Tools because it really has more inline as a full-fledged Task Manager alternative. It even has a UAC toggle in it to replace the default Task Manager so that seals it. I put it right next to Task Manager DeLuxe under Hardware monitors since that is listed as a Task Manager alt and fit nicely there (No need to create a new entry). I also decided to remove NeoHtop from being its own Task Manager alt entry and merged it with Task Manager DeLuxe and System Informer since I don't think it really needs to be seperate on it's own.

The newest addition made is that I added an entry for a Windows 7 Task Manager that can be used for Windows 10/11. I have tried it and it worked very well in my experience, what-with an option to replace the default Task Manager in Windows 11 and the program being extremely efficent and lightweight since it's literally just the Windows 7 TM (Along with an option to replace the new msconfig.exe with the old one that allows management of startup apps before it got removed). I think some people would enjoy it so I think the addition fits well there. FMHY also lists other old Windows stuff like RetroBar under Customization so it fits with promoting old Windows stuff for new Windows OSs.


nbats:
thank you the changes look fine to me, I will apply them all soon, but we will have to test the win 7 task manager ourselves before adding, so I moved that to our discord

PR #4991, opened 2026-03-24
Updated italian section

PeppeMonster:


nbats:
thank you for doing this, I am editing now so I'll merge these changes into that now

PR #5002, opened 2026-03-26
I've been using openanime for a while and I'd like to see it starred here.

chunyt:
It's safe to use and it's pretty fast, works on mobile too.


nbats:
fair enough thanks for the suggestion, I'll merge this into main edit later on

PR #5003, opened 2026-03-26
Add oyunindir.vip for Turkish

chunyt:
it's more like downloadpirate.com but for games and software, the reason I've starred is I've been using it for years and haven't got any virus but you definitely need an adblocker, site has some misleading ads.

### DIFF
--- a/docs/non-english.md
+++ b/docs/non-english.md
@@ -1721,6 +1721,7 @@
 
 ## ▷ Downloading / İndirme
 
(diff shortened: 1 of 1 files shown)


nbats:
the issue we had with it is they don't properly credit the sources of their games, which makes it hard for people to trust them, but thank you for sharing

chunsal:
yes that's right they don't give credits but they use trustful sources like fitgirl or m0nkrus, they check the program if it's legit or not then reshare it without credits. It's not ethical but it's safe.

PR #5016, opened 2026-03-27
Added APK Stores and fixed a streaming service url

chunyt:
uptodown is already on the fmhy, added turkish language version. Added aptoide (works best on older devices and it's partly open source)


nbats:
Thank you but we don't normally add apk, software, or game download sites to the non-eng section, our advice is usually to use the main eng section, and just switch to the respective language you need. 

I will get the hdfilmcehennemi url updated.

PR #5034, opened 2026-03-30
Update privacy.md (a Google keep list of scanners I use for 0 day threats)

AmadeyHunter:
changes at file changes tab + notes from me, it all of them are free and have like no imits besides their tos and file upload limit (which I forgot to list)


nbats:
thanks for sharing with us, I'll pass all of these along to our discord for testing

PR #5035, opened 2026-03-30
Add Fancy font generator to text tools list

vaishnav-v:
Add Fancy font generator to text tools

site url: [font generator](https://givemefonts.com/)

Used for creating unicode fonts


nbats:

PR #5042, opened 2026-03-31
add: Elephant to YouTube Downloaders

Margeshy:
Paste yt link (can be from url or share link) and elephant routes yt-dlp into the FDM gui


nbats:
thank you but we have it listed next to freedownloadmanager itself already

Margeshy:
> thank you but we have it listed next to freedownloadmanager itself already

Yeah I know but JDownloader has been listed in the wiki 3 times (4 if you count beginner's guide) so I thought to suggest this as well
Elephant is actually pretty neat for those who want the power that yt-dlp provides yet the simplicity of a gui

PR #5050, opened 2026-04-02
Removed GitHub Android link in iOS Adblocking for AdGuard

nqlp:
Removed GitHub Android link in iOS Adblocking since it does not belong there.


nbats:
ty will fix

PR #5060, opened 2026-04-03
temp fix

bbbreaddd:
API backend is currently broken so I added this as a temporary fix to prevent people from just spamming feedback.


bbbreaddd:
Nevermind, I think I found the fix

PR #5064, opened 2026-04-03
Add AMWin-RP, Apple Music Rich Presence with Last.fm Scrobbling

jijirae:
AMWin-RP is a Windows 11 app that allows for displaying Rich Presence for Apple Music on Discord, with Last.fm scrobbling. Music Presence is currently testing this, but in beta so limited slots. AMWin-RP does this out of the box.


nbats:

PR #5080, opened 2026-04-05
Remove UC.Files from filehost section

Union-Crax:
UC.Files was too good to be true, and will turn invite only. Files can still be downloaded for a few days.


nbats:
alright sorry to hear that, I will get it removed next update soon

PR #5094, opened 2026-04-07
Remove ROM Heaven CSF from gaming tools list

helpme970:
Removed 'ROM Heaven CSF' from the list of gaming tools beacuse it already exists in Games > Download Moved Nemirtingas Epic Emulator from Client Tools to DRM Bypass


nbats:
thank you, I have editor open, so I will combine into that now

PR #5116, opened 2026-04-10
Change Google AI Studio & LMArena Listings w/ Updated Info

TimeAndLocation:
Some info updates.

- Arena AI requires a login to prompt any one of its models, meaning that you cannot simply reset the limits through a cookie reset. Instead, you need to switch your account to reset your limits.
- Google AI Studio can reset its limits through switching Google accounts, which I decided to add for clarity. I've also added a link to a Chrome extension that provides lag reduction for long chats; Google AI Studio's performance tends to be poor for anything more than light use. I personally use the extension, but I would suggest fully vetting it. That is, regardless, standard practice.


nbats:
I mean whos gonna make google accounts over and over? Seems unrealistic, but we will check out the new stuff added.

PR #5126, opened 2026-04-11
Updated links from SteamAutoCracks

Felidanger:
Updated links from SteamAutoCracks as they were taken down off github earlier this month

replaced with gitlab


nbats:
thanks for the heads up, I have editor open, so I'll merge into that now

PR #5129, opened 2026-04-12
Fix: Remove Broken SSL Sites

ibnaleem:
Removed a few sites that have broken TLS configurations.


eli32-vlc:
Hello,

I have been unable to replicate this issue on my end. Please verify that your clock is set correctly and that you do not have any TLS MITM software or proxy running.

All of these sites are functioning as expected. If you require further assistance, please join the FMHY Discord.

Thank you.

PR #5135, opened 2026-04-13
Updated the requirements for Claude

MathewJake365:
Updated the requirements for Claude, as it requires a phone number regardless of whether you sign up with email or Google.


nbats:
thanks for the heads up, I'll get it adjusted soon

PR #5140, opened 2026-04-14
Some small cleanups

peternmuller:


nbats:
This does not seem "small." We need some form of explanation as to what this edit is... 
image

Overimagine1 (changes requested):
> It has been taken down

No it didn't

nbats:
We appreciate you running the commands, but its just too much at once right now, if we ever want to run them ourselves in the future we will.

peternmuller:
> This does not seem "small." We need some form of explanation as to what this edit is...

Yes, sorry, I did this pull request before doing that `npm run format` commit and forgot to update it.
https://github.com/fmhy/edit/pull/5140/commits/dae6a1dca7445909122b0ce7e1f318efd23e7705 is still good to pick, though.

> > It has been taken down
> 
> No it didn't

Your're right, it's actually just in France that it's down:
image
https://lumendatabase.org/notices/46649330

PR #5146, opened 2026-04-15
Update text-tools.md

sierregarcia:


nbats:
thanks for sharing, we'll go through the list when we can

PR #5148, opened 2026-04-15
Update gaming.md

Margeshy:
Star RexaGames: Personally better than SteamRip. Large library + fast updates on cs.rin and site + modding is way easier as there are no license check errors
Position could be changed


nbats:
There's a thread going for it currently in our Discord, we'll see how people feel there.

PR #5162, opened 2026-04-17
Add and remove French streaming websites

MicheDePainn:
I removed Yakriv and Deksov, which had expired, and added Dotriv


nbats:
thanks for letting us know, we'll get links fixed next update

PR #5170, opened 2026-04-18
Update video.md

kisskh9:
Added kisskh.red as an additional mirror for KissKH. This ensures users have an alternative access point in case the primary domains are down or blocked in certain regions. Tested and working.


nbats:
thanks for the heads up, I have editor up, so I will paste this change into that

PR #5175, opened 2026-04-20
Fix spelling ubo to uBlock Origin

nqlp:


nbats:
thank you will adjust soon

PR #5188, opened 2026-04-22
Add magnetz to torrenting

anoncry:


nbats:
thanks for sharing, I re-opened the discord thread we had going for this so we'll see how people feel

PR #5221, opened 2026-04-28
Add Groq Whisper to Audio Transcription

arthiccc:
Added Groq as an alternative WebUI for Whisper transcription in the Audio Transcription section.

## Changes
- Added [Groq](https://console.groq.com/playground?model=whisper-large-v3-turbo) to the Whisper line


nbats:

PR #5245, opened 2026-05-02
Update BetterRYM extention to EvenBetterRYM fork.

Theboombringer1:
The BetterRYM extension has not been updated since 11 Dec 2024, and some of its enhancements (ex: finding missing stream links) are broken. This fork fixes those enhancements and adds several more on top.


nbats:
thanks for letting us know, we'll get it switched next update to the better one

PR #5246, opened 2026-05-02
Add recently-removed page

bbbreaddd:
## Preview: https://4d0907ac.edit-b69.pages.dev/

This adds a new page called "Recently Removed". The purpose of it is (obviously) to see the recently removed sites from the last 30 days. I felt like this was a good idea, because if someone is searching a site that was removed, they can see it was removed rather than being confused why they can't find a specific site.
It could be longer than 30 days if needed, I just felt 30 days was good enough.
The Recently Removed page is generated on build, so it's not something you commit to or view inside the repo.
You can view the page in the "More" section.

I also added a new feature that lets you exclude a certain section from being added to the search index.
## Example:
```md
- [Site](URL) <!-- search-exclude --> - Description (Removed in hash) <!-- /search-exclude -->
```

This was used to exclude the descriptions in the Recently Removed page from the search index, but it could also be used in other pages as well. 
> [!WARNING]
>  Maybe it would be best to keep it exclusive to Recently Removed for now since other sites like Reddit, single-page, or Rentry probably can't handle it properly.

image

image


Margeshy:
Really cool. Would be better to move it from under the Move section where it's hidden, maybe below "Posts"?

PR #5253, opened 2026-05-03
Added Mobiflix

Golphin0:
I've added Mobiflix to the Android Streaming category, which is essentially OnStream but without the need of an external player, like SPlayer.


nbats:

PR #5256, opened 2026-05-04
Remove Create React App from developer tools

NandkishorJadoun:
Removed Create React App entry from developer tools list because it is is deprecated. 

https://react.dev/blog/2025/02/14/sunsetting-create-react-app


nbats:
thanks for the heads up, we'll remove it soon

PR #5311, opened 2026-05-14
[ci] update actions to support node 24

Project516:
https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/#removal-of-operating-system-support-with-node24

### DIFF
--- a/.github/workflows/deploy-api.yml
+++ b/.github/workflows/deploy-api.yml
@@ -17,7 +17,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [20]
+        node-version: [24]
 
     steps:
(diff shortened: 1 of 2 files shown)


nbats:
image

Project516:
@nbats Workflows like checkout v4 use node 20, that is why i am updating them to versions which support node 24

base-13:
> @nbats Workflows like checkout v4 use node 20, that is why i am updating them to versions which support node 24

fair

base-13:
though has the workflows been tested with these versions?

rhld16:
> though has the workflows been tested with these versions?

had a quick skim of the updates for all the actions - nothing majorly breaking.

PR #5330, opened 2026-05-16
remove ancient chatgpt widescreen userscript

jackmhny:


nbats:
ty we will remove it soon

PR #5331, opened 2026-05-16
Adding UniFab as a video upscaler/enhancer

lasonmason:
Previous commit got lost and so did my github account :(

I am adding unifab as another application for people to use (a crack is available on lrepacks), instead of only topaz

also unsure why the random link at the bottom was edited


nbats:
ty we'll get this added next update

PR #5332, opened 2026-05-16
Add FastScrobbler to last.fm tools

suotn:
FastScrobbler is an Open Source, lightweight scrobbler for Apple Music on Apple devices that works more consistently than its alternatives. Includes one-time purchasable, non-essential Pro features


nbats:

Issue #5326, opened 2026-05-15
Declare the RSS feed on the whole site


NotaInutilis:
RSS is cool! Currently there's only one link to the feed on the whole site, and it's on this page https://fmhy.net/posts

So it's not picked up automatically by feed readers when submitting the main https://fmhy.net URL and only the posts page triggers the automatic detection process.

If I'm not mistaken (I can be), there should be a link to the feed in the site's `<head>` so in order for it to be properly declared and recognized by feed readers.

saiuttejr:
should be fixed by #5327, added the missing link in the head

nbats:
thanks for the suggestion, I made a thread for it on our discord so we'll see what happens

PR #5338, opened 2026-05-17
add rss detection

bbbreaddd:


bbbreaddd:
For https://github.com/fmhy/edit/issues/5326

PR #5350, opened 2026-05-19
Add French Stream to non-english.md

GillesLab:
It’s a very good website; the video quality isn’t always great (max 1080p), but that makes the site accessible to older devices.


nbats:

PR #5351, opened 2026-05-19
search overhaul

bbbreaddd:
I don't think it's fully ready, but I'm hoping I could get feedback and testing since it's hard to test everything.
Here are the new features

- Search History
- Search suggestions: If there are no results but your search is close to another word, it will suggest it to you. For example, if you search up pihole, it will suggest pi-hole. If you search up monkrus, it will suggest m0nkrus
- Show More Results button: Previously, search results were limited to 16 results only. Now, you can just press "Show More Results" and it will load more results.

Smaller stuff:

- Loading Circle: When a search is loading, it will just show that it's loading rather than just doing nothing
- Possibly faster searching? It seems like it is faster.
- It will now show how many matches the search found
- Removed webview clear button
- Fixed formatting in search preview (bold fonts)
- Excluded blog posts older than 2 months
- Small UI changes

### DIFF
--- a/docs/.vitepress/constants.ts
+++ b/docs/.vitepress/constants.ts
@@ -38,6 +38,22 @@ export const search: DefaultTheme.Config['search'] = {
         return ''
       }
 
+      // Exclude posts older than 2 months (60 days)
+      if (relativePath.includes('posts/')) {
+        const frontmatterMatch = src.match(/^---\r?\n([\s\S]*?)\r?\n---/)
+        if (frontmatterMatch) {
+          const dateMatch = frontmatterMatch[1].match(/^date:\s*['"]?([\d-]+)['"]?/m)
+          if (dateMatch) {
+            const postDate = new Date(dateMatch[1])
+            const twoMonthsAgo = new Date()
+            twoMonthsAgo.setMonth(twoMonthsAgo.getMonth() - 2)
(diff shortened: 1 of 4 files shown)


bbbreaddd:
Preview: https://c36e93a5.edit-b69.pages.dev/
- Added scrollbar gutter to prevent the background moving when search is open
- Cached last search results after closing the search bar
- Fixed slowness when scrolling through many search results
- Faster same page scrolling
- Increased debounce (so it should lag less when typing fast)

bbbreaddd:
Update: <https://26ffc35e.edit-b69.pages.dev/>
- Better search sorting. It will now check for things like if a site is starred, if the search query is an exact match, if the search query is a prefix of a word, if the search query is a link (rather than the description) and it will prioritize things based on these factors
- More sensitive suggestions
- Ignore invisible zero-width characters (which seem to be everywhere in the wiki for some reason)

bbbreaddd:
Rebased to resolve conflicts.
- Cleaned up dead code
- Removed negative scoring
- maybe optimized some stuff
- changed the alignment of the platform and tooltip icon because it wasn't really aligned that well which was more obvious in the search preview

bbbreaddd:
Welp, it's been long enough. I think it's ready. 
Preview: https://3a9b1cff.edit-b69.pages.dev/

PR #5353, opened 2026-05-20
Add WeFlix to streaming site list

kweephyo-pmt:


nbats:
it seems to be a frontend for some common apis, we get sent many similar, it needs to stand out more sorry

PR #5357, opened 2026-05-20
theme improvements

bbbreaddd:
Stops the default theme from showing when loading the site and a custom theme is selected
Improved outline thing on mobile so it wouldn't take up the entire screen anymore


bbbreaddd:
Before: 

https://github.com/user-attachments/assets/d2262bec-9bee-4e0d-a9d2-804558666dff

After: 

https://github.com/user-attachments/assets/5963292c-48df-4503-8dbc-033559579798

PR #5360, opened 2026-05-20
Static Singlepage

bbbreaddd:
Recreated the single-page but on build, so it will be accessible on https://fmhy.net/single-page.md instead of https://api.fmhy.net/single-page

Preview: https://99701493.edit-b69.pages.dev/single-page.md

It shouldn't be that big of a change
- It should be faster
- Won't be rate-limited
- Won't rely on Cloudflare Workers (at least I think that's what it uses)
- It's generated when the wiki is built rather than generating all the files from the Github, so it should be more accurate

We could make a redirect from the old URL to the new URL, or we could just leave both accessible. Either way, that would need to be done in Cloudflare which I don't have control over


nbats:
I think leaving both accessible is probably best, I think we have a fair amount of stuff that relies on the api address if I'm not mistaken, like our bot and probably other stuff.

PR #5367, opened 2026-05-22
fix: close theme VDropdown when clicking outside

eli32-vlc:
Fixes an issue where clicking the theme icon then another nav item (e.g. Ecosystem) leaves the theme dropdown visible, causing it to overlap the other menu.


eli32-vlc:
Closing as requested

PR #5390, opened 2026-05-25
Add Grokipedia to the list of encyclopedias

degodfather:
Added a new entry for Grokipedia, an AI powered encyclopedia.


nbats:
it was rejected multiple times sorry, people don't feel using a AI for information like this is a good call

PR #5404, opened 2026-05-26
added duck ai

MrCook284:
i wanted to add Duck.ai as i as thought of it as a alternative to Ask Brave with a bit more flexibility because you can choose what model to use. I don't use ai enough to know of any rate limits that it could reach.


nbats:
thank you but this has been discussed and rejected in the past a few times now for having bad limits

PR #5405, opened 2026-05-26
Add Hi-fi Audio Bot to audio resources list

creeny0:
Hi-Fi Audio Bot is a well-working lossless audio source, free (and you have to follow their channel) or paid (but not really necessary). IMHO it works so much better than everything else in that list


nbats:

PR #5406, opened 2026-05-26
old school fonts should be starred

MrCook284:
when I was going through this page I was able to find this website and I thought it was odd it didn't have a star. The amount of effort into previewing the fonts, organizing, and the ease of use of the website is very high. 

I don't know if it is determined though the discord but I though I would shoot my shot.


nbats:
thanks for the suggestion, I'll make a star thread in discord to see how people feel

PR #5408, opened 2026-05-27
Add anon mode toggle to search

eli32-vlc:
Adds an anon toggle button (eye/eye-off icon) next to the fuzzy search toggle. When active, searches are not saved to recent searches.

### DIFF
--- a/docs/.vitepress/theme/components/VPLocalSearchBox.vue
+++ b/docs/.vitepress/theme/components/VPLocalSearchBox.vue
@@ -1,5 +1,5 @@
 <script lang="ts">
-import { ref, h } from 'vue'
+import { h, ref } from 'vue'
 
 const RESULTS_PAGE_SIZE = 16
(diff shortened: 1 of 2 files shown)


nbats:
we'll take a look, I'm just unsure on this as people can clear their cookie so easily and achieve the same thing

bbbreaddd:
Made some changes to it. Preview: https://dd2ed360.edit-b69.pages.dev/
https://github.com/bbbreaddd/edit/tree/feat/anon-mode

- Saved the setting in local storage instead of resetting on reload
- Moved the button to a footer rather than being in the search bar. I feel like it fits better in here. I don't think it really needs a dedicated button on the search bar to be seen all the time.
- Cleaned up some code. Ignore the constants.ts changes, that was just Prettier bothering me about something. Don't know why it didn't before.

I tried to make a PR on your repo, but couldn't for some reason.

nbats:
gonna close this, looks like bread is going to make changes

PR #5414, opened 2026-05-28
feat: search scroll to match by manuJL :)

manuJL:
i added search-scroll-to-match feature to the wiki  and it was working for  me and test it and if its works for you guys than merge it


manuJL:
i used Claude code for this

PR #5418, opened 2026-05-29
Fix Cineby URL in video.md

parth-awasthi:
Changed URL from Cineby.sc to Cineby.cc

### DIFF
--- a/docs/video.md
+++ b/docs/video.md
@@ -10,7 +10,7 @@
 
 ***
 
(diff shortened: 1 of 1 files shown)


waitlists3:
why would it be cineby.cc and not cineby.sc?

parth-awasthi:
Because that url is not working and this is the new one

waitlists3:
no, that url doesnt work because you are from india, and its blocked in india, cineby.cc is a fake clone site. use a vpn to access cineby.sc

parth-awasthi:
Oh my bad mate ill close this PR.

PR #5437, opened 2026-05-31
Update BuffBook link to the official website

classy-giraffe:
Updated website from workers url to official url.


nbats:
thanks for the update, I just got it merged into the main edit earlier so it should show up on site soon

PR #5474, opened 2026-06-05
update mirror for ppv.to

boat1337:
changed mirror link from ppv.cx (dead) to ppv.is


nbats:
thanks for the heads up, I'll get it adjusted in next update

PR #5479, opened 2026-06-06
fix: correct formatting for SportsBite link in Live Sports section

nqlp:
Remove one parenthesis


nbats:
ty already fixed in editor will send soon

PR #5498, opened 2026-06-09
updated fighting games section

hhoneeyy:
removed sf6frames because it was shut down


nbats:
thank you I will merge these changes into current edit soon

PR #5502, opened 2026-06-10
fix fluffle.cc url

omansh-krishn:
https was missing 'h'


nbats:
thank you I'll merge this into editor now

PR #5509, opened 2026-06-11
Added Mercado Play

LucasDondo:
Mercado Play is a free w/ ads streaming platform built by the biggest company in Latin America, Mercado Libre. This makes it very reliable, secure and well-known.


nbats:

PR #5524, opened 2026-06-14
Ignore infobox if it matches in the search

manuJL:
worked on the ignore-infobox-if-it-matches-in-search
suppose If you search up `uBlock Origin,` it will show the note first before the actual link. here the live preview :https://exact-given-template-organised.trycloudflare.com


bbbreaddd:
This can probably be closed as it was already merged in a separate PR.

PR #5532, opened 2026-06-14
added monitor refresh rate test tool

abhranil26:
Frame rate testing tool was already there so added a refresh rate tester, will come in handy.


nbats:

PR #5540, opened 2026-06-15
Replaced a 404 link with a working link.

abhranil26:
https://tools.rodrigopolo.com/display_calc/ was 404 so replaced it with a link of a working tool and added a Refresh rate tester tool too!


nbats:
ty for sharing, I'll remove dead one + we'll test new one

PR #5562, opened 2026-06-17
Search by url and applied new tweaks and  changes also original was by lou 

manuJL:
original work was by Lou i just applied the following tweaks 
Tweak 2 — loose URL matching (fixed)
urlMatchesQuery now only matches when the stored URL contains the query, dropping the reverse queryValue.includes(urlValue) direction. So https://pi-hole.net/testttttt no longer matches https://pi-hole.net/. Exact and partial-path matches still work because equal/contained strings still match forward. Verified against your example earlier.

Tweak 1 — fuzzy/exact sort (Option C)
Within a tier, raw score stays the primary tiebreaker; added a deterministic fallback (title, then id) only for exact score ties, so equal-scored results no longer randomly reshuffle between modes or re-renders. Relevance ordering is unchanged.


bbbreaddd:
Oh forgot to comment on this.
Here is the preview https://22cb2cf9.edit-b69.pages.dev/

Here are my changes:
- Ignored specific domains to prevent noise in results. For example, when searching up `github`, it won't show *every* single Github repo in the wiki, but rather Github related links. 
- Search also focuses on things that are identifiers. Like for example, `github.io` is excluded as many people use github.io, but we still add the subdomain in the index so you can search up `privacyguides` and `https://privacyguides.github.io/` will show up.
- Starred links will rank higher (but not too high) as they rank differently than regular text.
- The search will automatically detect URLs and soft-enable URL searching, even when it's disabled.
- Higher character limit to account for longer URLs
- Changed URL Search icon
- Reduced the size of the search index by 33%
- Junk like `www.`, and `?foo=bar` and other query parameters are ignored as they are generally not very useful in the search
- Removed invite and URL shorteners from the search index as they are mostly random
- Capped the amount of URLs that can show up in results with only 8 per domain (60 total)
- Fuzzy tolerance is now dynamic with short queries matched more strictly

Misc:
- When the focus is on the search previews rather than the search box, typing will automatically bring you back to the search box.
- Some optimizations.
- Fixed some errors/problems that were popping up.

PR #5581, opened 2026-06-19
Add rentry link for UC mirrors

Union-Crax:
So fmhy contributers dont have to remove dead mirrors of UC when one goes down. Rentry will automatically update.


nbats:
thanks for the update, I'll merge it into the next update soon

PR #5593, opened 2026-06-21
Update reading.md to remove readarr

CannedToes:
Removed 'Readarr' from the list as it has been retired and isn't really worth using any more anyway, https://readarr.com/


nbats:
ty for the heads up, we'll remove it soon

PR #5595, opened 2026-06-22
fix: Magisk tools should be below masgisk entry

fynks:


nbats:
thanks for the heads up, will merge into main edit soon

PR #5601, opened 2026-06-22
Update privacy.md

stuffed18:
Starred Lantern VPN under Proxy. 
The speed limit after 250 mb has been removed and the vpn is much faster than before. It has also gotten a major UI/UX update. 

Additionally, this VPN works on networks that block all other VPNs.

### DIFF
--- a/docs/privacy.md
+++ b/docs/privacy.md
@@ -444,8 +444,8 @@
 
 # ► Proxy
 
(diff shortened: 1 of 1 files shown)


nbats:
I'll make a suggestion thread in Discord with what you said here and we'll see

nbats:
are you sure they got rid of the limit? says here they still have it from their site: 
image

stuffed18:
Yea its unlimited despite what the website says, there used to be a bandwidth limitation but not anymore.

PR #5620, opened 2026-06-25
Modify Votify audio quality details

nikopoto:
Updated Votify entry to reflect changes in audio quality options.


nbats:
thank you much appreciated, I have editor open now, so I'll just copy this into that now

PR #5631, opened 2026-06-25
some preview was not showing in the search results.

SarkarKurdish:
When detailed search view is on, some results show preview text under the title. But for sections like Stars Added ⭐ in posts, only title was showing — no preview body.

Example: search oceanofpdf shows Reading page preview fine, but Posts / Stars Added had no preview.

image

Why it happen
Search index and excerpt cache was using different anchor names for same section.

Index use slug like stars-added (from link href)
Excerpt cache was saving under stars-added-⭐ (from heading id)
So cache lookup fail > empty text > no preview.

Also encoded anchors (like %E2%AD%90 for ⭐) was not decoded same way on both sides.

now:

Decode section anchors before cache read/write
Save excerpt under both href anchor and heading id when they different
On lookup, if exact match fail, try prefix match (so stars-added find stars-added-⭐)

result

image

### DIFF
--- a/docs/.vitepress/theme/components/VPLocalSearchBox.vue
+++ b/docs/.vitepress/theme/components/VPLocalSearchBox.vue
@@ -691,9 +691,7 @@ watch(
       if (canceled) return
 
       const mapped = candidates.map((r) => {
-        const [id, anchor] = r.id.split('#')
-        const map = cache.get(id)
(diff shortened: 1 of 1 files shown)


nbats:
was fixed here https://github.com/fmhy/edit/pull/5641

PR #5637, opened 2026-06-26
adding my list page

SarkarKurdish:
image
image

I've been using FMHY as a reference and kept losing track of links I meant to come back to.

**How it works**: when you hover a link on any wiki page, a small Bookmark button shows up next to it. Click it and that row gets saved to localStorage.

When rendering My List, the app tries to re-fetch the live row from the original page so you see the current version with tags and formatting intact. If an entry was removed from the wiki, it falls back to the HTML snapshot saved at bookmark time and marks it as removed. You can also remove items from My List with the X button.

There's a "My List" link at the top of the sidebar. Bookmark buttons don't show on the My List page itself (would be weird to bookmark bookmarks).

This is a draft and I'm definitely open to feedback, naming, UX, where the button appears, whether sidebar placement is right, performance of loading saved pages in the background, all of it. Happy to adjust before this is ready to merge.

### DIFF
--- a/docs/.vitepress/shared.ts
+++ b/docs/.vitepress/shared.ts
@@ -36,7 +36,8 @@ export const excluded = [
   'feedback.md',
   'index.md',
   'sandbox.md',
-  'startpage.md'
+  'startpage.md',
+  'my-list.md'
 ]
 
 const safeEnv = (key: string) =>
@@ -112,6 +113,10 @@ export const nav: DefaultTheme.NavItem[] = [
 ]
 
 export const sidebar: DefaultTheme.Sidebar | DefaultTheme.NavItemWithLink[] = [
+  {
+    text: '<span class="i-twemoji:bookmark"></span> My List',
+    link: '/my-list'
+  },
   {
     text: '<span class="i-twemoji:books"></span> Beginners Guide',
     link: '/beginners-guide'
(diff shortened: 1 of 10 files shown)


bbbreaddd:
I'll have to look at this later, but I'm wondering, this is desktop only, right? I'm guessing there isn't any mobile support.

SarkarKurdish:
> I'll have to look at this later, but I'm wondering, this is desktop only, right? I'm guessing there isn't any mobile support.

There is, in mobile view the bookmark button apears after all the icons (discord, …etc)

nbats:
I'm not sure I follow this.. can't people just use bookmarks, we have a entire section for it

also we don't have accounts so people will just lose all their bookmarks anytime they clear cache

bbbreaddd:
Well, it would be stored in the browser storage, so it will not sync between different devices, but it won't delete unless you delete your site data (cache is separate). And sure, people can use other bookmarks, but I don't know, maybe people might want to save it in FMHY. I mean, if you want my opinion on it, I likely wouldn't use it and I don't like how it's currently implemented, mainly due to the UI. I don't have any ideas on how else to implement it though.
I'm not completely against it, I just want it's implementation to be intuitive and for it to work well.

nbats:
I just don't think its worth considering unless it has some way to sync across devices.. but idk if its worth that much effort when countless bookmark apps exist or even just browser bookmarks

if people feel very strongly about it a #suggestion thread can be made in discord instead and we can work on it from there rather than a single pull request

PR #5648, opened 2026-06-27
Update audio.md

ghost:
squid.wtf doesn't support Qobuz and Tidal anymore. It's confirmed by the devs in their discord. "Qobuz has blocked the account again; in my opinion, there’s no point in keeping Qobuz going any longer. Qobuz is Dead" this is their official announcement.


nbats:
thanks for the heads up, I have editor open now, so I'll just copy this change into that now for next update

PR #5682, opened 2026-07-01
Change focused search  and scroll-to-match highlights to match themes.

litekin:
This changes the currently focused search and scroll-to-match highlights to use existing theme colors with some tweaks, rather than dark yellow across all themes.


bbbreaddd:
Looks good

PR #5693, opened 2026-07-02
Bug fixes and cleanup

bbbreaddd:
Preview: https://af3b37a3.edit-b69.pages.dev/
(although not much has changed that is visible)

- Fixed post dates to use a specific timezone rather than local to show consistent and accurate dates.

Before:
brave_9s3GRXYASk

After:
brave_dPe6B6Ckv0

- Fixed invalid HTML for icon tooltips
On load, the page shifts around due to invalid HTML with the tooltips, now it's stable.

- Stopped precaching images/search index in PWA

- Optimized the June theme

- Fixed a bug in the recently-removed stage that bloated the gitconfig with duplicate lines. Not harmful, but annoying.

- Small cleanups

### DIFF
--- a/docs/.vitepress/config.mts
+++ b/docs/.vitepress/config.mts
@@ -149,7 +149,7 @@ export default defineConfig({
             var today = new Date();
             if (today.getMonth() === 5) {
               document.documentElement.classList.add('june');
-              function applyJuneTheme() {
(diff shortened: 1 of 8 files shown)


nbats:
seems to have conflicts

bbbreaddd:
Fixed

PR #5696, opened 2026-07-02
Lint Markdown Cleanup

bbbreaddd:
Just did a lot of small changes like capitalization, dashes, making sure things are consistent. Nothing that important. You can look at the commit names to have a better idea of what I did.
One controversial thing I did though was renamed all "Mac" references to "macOS". I feel like it wasn't that big of a deal and I feel like macOS is a better way to reference mac OS than just Mac. I also made sure both Mac and macOS were supported by the site though so technically either can be used.


nbats:
has conflicts

bbbreaddd:
Fixed

Issue #5711, opened 2026-07-04
lint-markdown a/an check misses consonants g, h, j in Check 12


chirag127:
## Problem
`scripts/lint-markdown.js` Check 12 ("Incorrect article 'an' usage") fails to flag `an` before words starting with `g`, `h`, or `j` because those letters are missing from the consonant character class.

## File:line
`scripts/lint-markdown.js:496`

```js
const anAMatch = line.match(
  /\b(an)\s+([bcdfVkLmMnNpPqQrRsStTvVwWxXyYzZ]\w+)/i
)
```

The class contains b c d f k l m n p q r s t v w x y z but omits g, h, j.

## Steps
```sh
node -e "console.log('an game console'.match(/\b(an)\s+([bcdfVkLmMnNpPqQrRsStTvVwWxXyYzZ]\w+)/i))"
node -e "console.log('an javascript lib'.match(/\b(an)\s+([bcdfVkLmMnNpPqQrRsStTvVwWxXyYzZ]\w+)/i))"
```

## Expected
Both match (flag as incorrect article use).

## Actual
Both return `null`. `an banana` correctly matches, confirming the class is the issue.

## Environment
node ^22.13.0 (per `package.json` engines), Windows 11. Reproducible on any node.

Fix: add `gGhHjJ` (or use `/i` + `[bcdfghjklmnpqrstvwxyz]`).

Thanks for maintaining fmhy/edit!

nbats:
thanks I'll pass it along to devs

PR #5729, opened 2026-07-06
Improve markdown linter and add pre-commit bypass

bbbreaddd:
Markdown linter:
- Fix the a/an grammar check (it missed words starting with g, h, j) https://github.com/fmhy/edit/issues/5711 and stop it wrongly flagging cases like "an hour", "a European", and "an m3u"
- Ignore the posts/, other/, and public/ folders, and skip non-file paths
- Only use colors in a real terminal, so captured logs (VS Code, CI) no longer show raw color codes like [36m
- Underline the exact part of the line that has the problem, making issues easier to spot

Pre-commit hook:
- When a check fails in a terminal, ask whether to commit anyway; when not interactive (IDE/CI), show how to bypass with `git commit --no-verify`
- Quiet down lint-staged's output so failures are cleaner to read

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "prepare": "simple-git-hooks"
   },
   "simple-git-hooks": {
-    "pre-commit": "pnpm lint-staged"
+    "pre-commit": "sh scripts/pre-commit.sh"
   },
   "lint-staged": {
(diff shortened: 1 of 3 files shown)


bbbreaddd:
and improved the message that shows up when Prettier fails as well

PR #5737, opened 2026-07-07
Add feedback blocklist to prevent toxic messages and spam of profanity-

manuJL:
feedback-blocklist.txt its simple text file that has list of vulgar and toxic and profanity words  that is used by feedback.post.ts   to detect those words and blocked them and  feedback with those words don't get sent instead it get discarded  and regex patterns  and  keeps the original rate limiting and Discord webhook logic

### DIFF
--- a/api/feedback-blocklist.txt
+++ b/api/feedback-blocklist.txt
@@ -1,0 +1,57 @@
+# FMHY Feedback Blocklist
+# One word/pattern per line. Lines starting with # are ignored.
+# These are checked server-side; toxic feedback is silently discarded
+# and never reaches Discord.
+
+# Racial slurs
+nigger
+nigga
+n1gger
+n1gga
+faggot
+fag
+faggit
+fagg0t
+tranny
+spic
+sp1c
+chink
+ch1nk
+kike
+k1ke
+coon
+c00n
+wetback
+gook
+kaffir
+darkie
+
+# Gender/sexist slurs
+bitch
+b1tch
+betch
+whore
+slut
+slutt
+cunt
+c0nt
+
+# General insults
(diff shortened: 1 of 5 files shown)


manuJL (commented):
added small change

litekin:
I'd suggest using https://www.profanity.dev/ or importing everything in https://github.com/dsojevic/profanity-list, rather than manually choosing which words to block

bbbreaddd:
I have a few problems.
1. You changed the copyright header for some reason. You weren't the original author of the file so I don't think it's right to do that.
2. With the way it is right now, it seems likely for false positives to happen. I think it should be more specific to ensure that only likely toxic messages are blocked. It shouldn't be too broad.
3. You say that blocked messages are indistinguishable from a regular message, but it isn't. For some reason, you add to the message if it's blocked. Technically most people wouldn't know the difference, but still, the difference is there and I'm not sure why it would be there.
4. Since it's open source, it probably wouldn't be that hard for someone to just look at the code to see how to workaround the filter, but again, most people probably wouldn't know how to do that or know about the filter.
5. I'm not really liking the idea of blocking messages (mostly because of false positives). I don't really care when I see toxic messages, it's easy for me to ignore them and delete them. Not sure if other people feel the same way.

I'll see if I can test it later on to find out if it actually works.

litekin:
yeah this would just be something to reduce trashy spam. better than nothing.
currently it seems you just had grok generate a list of profanities based on https://github.com/dsojevic/profanity-list... rather than importing the list. use the npm package instead. then the JSON properties can be used to minimize false positives, by setting allow_partial to false and the severity level to 3 or 4.

PR #5741, opened 2026-07-08
fix: Updated Burning Series Links

SnowJuli:
Updated link for Burning Series to include new and alternative domain.


nbats:
thank you will merge into editor next update

PR #5748, opened 2026-07-09
small fixes

SomebodyDoingThings:
fix 1 (internet-tools): remove the 2nd stoat resources link as it is a redirect to the first one

fix 2 (feedback): add 1 missing quotation mark


nbats:
ty will merge both changes into our next update

PR #5757, opened 2026-07-10
Removed distro.moe, as the domain has expired

KalerKaler:
distro.moe is not working anymore, hence removed it from the page


nbats:
ty we'll remove next update

PR #5767, opened 2026-07-10
fix: correct spelling of "Mirros" to "Mirrors" in Live TV section

nqlp:


nbats:
ty will merge this into the next edit

PR #5771, opened 2026-07-11
Add Insta Scope 

SoaibHaque:
- View instagram account
- Load more instagram posts and reels
- View and Download posts and reels
- Download user DP
- Download instagram video using URL


nbats:

PR #5781, opened 2026-07-12
Add Mastodon and BlueSky icons

klenndal:
Added icons for Mastodon and BlueSky to replace platform links, to match the style of other platforms


litekin:
nice

PR #5783, opened 2026-07-13
Feedback Improvements

bbbreaddd:
Preview: https://99fc1f42.edit-b69.pages.dev/privacy

- **Select a line from the page** - users can click the "Select a line from page" button and then click on any line to auto-quote it into their feedback message.
- New field for users to leave contact information
- **Translation** - Non-English feedback is now detected and will automatically be translated to English
Uses Cloudflare Workers. Also shows the original message.
- Client-sided rate limit (as a fallback for Cloudflare)
- UI Improvements
  - Button to send another message
  - Moved Back button to top left
  - Text box will now auto-resize automatically 
  - Discord feedback messages will now have colors for each category
  
Some of these features *might* be a little too much so I would like opinions, but I think they're useful.

image
image
image


bbbreaddd:
Oh forgot to mention that I also made a small unrelated improvement to pnpm format. It will now only check and modify files that are staged so it won't modify any files you aren't working on.

litekin:
looks great. changes I'd make:
1. make the example text in the text boxes the same: 
image

2. I feel like it would be better if the box that shows the quoted line *was* the button to select a line. you'd click it, then it would turn from the button into the quote, with a button to re-select/delete on the right side of the quote box or shown when hovered.
3. personally I would keep "Your feedback is valuable and helps us make FMHY better."
4. the big message like "We're always looking for ways to improve!." or "Nix the fluff and just tell us what you think!" seems to be random? I'd change that to always the same thing. and I'd just make it "We want to hear your feedback!" or something.

PR #5818, opened 2026-07-18
fix: update note formatting for Reddit video downloads

nqlp:


nbats:
ty will adjust it next update

PR #5821, opened 2026-07-18
Update SportOnTV link to new domain

sportontv:


nbats:
ty will merge into next update

PR #5845, opened 2026-07-20
chore: remove winmoji because of rerouted url to gambling site

farmatenl:
**remove winmoji because of rerouted url to gambling site**


nbats:
thank you, will remove it next update

PR #5861, opened 2026-07-22
edit Feedback widget

litekin:
- added email as a contact option (Kai approved)
- changed the example feedback input field to be sans font instead of monospace, like the example contact input field
- removed some overly silly and unnecessary prompt texts (sorry lol)
- changed `ex.` -> `e.g.` I just think that's better but it's personal preference ig


bbbreaddd:
Looks good

PR #5875, opened 2026-07-24
Remove Tokuzilla as they appear to have been shut down

LovekovGaming:
Both their main site and their backups have been down for over a week. Members of the tokusatsu community are speculating this to be part of a broader crackdown on tokusatsu streaming, as several other tokusatsu streaming sites were also taken down over the past few weeks.


nbats:
ty its removed now

PR #5881, opened 2026-07-25
Add getindevice to Social Media Tools

rishujay:
Adding getindevice all-in-one media downloader


nbats:
sorry we just get sent so many of these, its tough to convince people things like it are needed

rishujay:
Hi @nbats,

Totally understand! The main difference is that our site is 100% free of popup ads, forced redirects, and clutter, making it completely safe and simple for users.

It’s designed as a fast, clean, all-in-one alternative to the ad-heavy downloaders out there. Would love it if you could give it a quick look whenever you get a chance! 🙌

nbats:
sorry it got rejected, here was the reasoning 
image

PR #5892, opened 2026-07-27
Improve search results

bbbreaddd:
- Fuzzy mode will no longer be used when searching URLs preventing bad results (ex. https://zeigames.com matching to ".com" plain text)
- Improved detection of URLs so it doesn't falsely detect non-URLs as URLs as break search (ex. "Team V.R" not matching due to the period)


bbbreaddd:
Preview: https://38374d9a.edit-b69.pages.dev/

PR #5894, opened 2026-07-27
Remove PollCode as they have closed

koviubi56:
Remove https://pollcode.com/ from Storage > Poll Sites, as they have closed. See https://web.archive.org/web/20260713203938/https://pollcode.com/


nbats:
thanks for the heads up, will remove it next update

PR #5898, opened 2026-07-27
chore: remove ARMGDDN Browser from Game Repacks as they only offer VR games now

Star123451:
Since July 12th they have removed non-VR PC games from their offerings due to the steep costs of hosting them and VR games getting much more traffic. The site should be updated to reflect this.


nbats:
thanks for the heads up, we'll remove it next update

PR #5903, opened 2026-07-28
Add MultiDrive to system tools list

mokosiy:
MultiDrive is a free app without ads and nag screens that allows:
- Clone disks byte-to-byte
- Create full drive backups in RAW and ZIP formats, restore them 
- Erase disks with specified HEX pattern
- Start multiple tasks in parallel, if you have a bunch of drives to process


nbats:

PR #5920, opened 2026-07-30
Update Video Tools (Streaming) and add LuffyFiles to File Hosting

fileada:
# Changes

* Updated the **Video Tools** section for streaming.
* Added **LuffyFiles** to the **File Hosting** section.

## LuffyFiles

* Max file size: **5 GB**
* File expiry: **30 days** (auto-deletes after **30 days with no downloads**)
* No signup required


nbats:
thank you we did get luffy added earlier

PR #5921, opened 2026-07-30
Update file-tools.md - add: autofetch.de

autofetch-de:
Disclosure: I am the developer and operator of autofetch.

I placed it under Download Managers because it seemed like the closest existing section, but it is more accurately described as download automation. The web service monitors configured sources and creates matching jobs, while a local client downloads the files directly to the user’s device.

Please feel free to move it if another section is a better fit.


autofetch-de:
feel free to move this into another listing. autofetch is different than an irc-downloader or download manager...

nbats:

PR #5928, opened 2026-07-31
Add section filter chips to search results

Anarchydr:
Added category filter chips to the local search modal so users can narrow down results to a specific wiki section (like Gaming, Audio, or Reading).

Once you type a query, filter chips showing live match counts pop up under the search bar. Picking a section filters the results, "All" resets it, and your selection is remembered in `localStorage`.

A few notes:
- Categories are generated directly from the sidebar, so any sidebar renames update automatically. If Wiki and Tools share a page (like `/audio#audio-tools`), they get grouped under Wiki.
- Filtering happens on the in-memory results before pagination. Switching categories doesn't re-trigger the search index query, which keeps highlights, counts, keyboard navigation, and "Show more" working smoothly.
- If a section turns up zero matches, it shows `No results in <section> — Show all sections` so it's clear the search itself didn't fail.
- Ran `pnpm docs:build`, `pnpm lint`, and `prettier check`
- Tested filtering manually across exact, fuzzy, and URL search modes


flowerey (approved):
LGTM if it works. It is good that you didn't hardcode the chips :)

PR #5931, opened 2026-07-31
Move AppManager to the correct category

TheUltraParanoid:
Move AppManger from **Software Sites** to **Linux Apps** next to AppImageLauncer and GearLever as it fits there better.


nbats:
thank you, we'll merge this change into editor now

PR #5944, opened 2026-08-02
small edits

litekin:
- edited feedback prompt ->
> Please read the [contribution guide](https://fmhy.net/other/contributing) before submitting your feedback, and be sure to include a link if you want something added!

- changed mentions of https://fmhy.net/other/contributing to "contribution guide" instead of "contribute guide"
- fixed capitalization in gaming download note
- edited https://comparisontabl.es/ description and ranked higher

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,4 +1,4 @@
-# Contribute Guide
+# Contribution Guide
 
 > [!INFO] NOTE
 > Some of these steps are easier if you're in our [Discord](https://github.com/fmhy/FMHY/wiki/FMHY-Discord). It opens every Friday.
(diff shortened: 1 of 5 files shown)


flowerey:
@litekin "edited https://comparisontabl.es/ description and ranked higher" was this discussed? like you can bump but there should be a reason

flowerey (approved):

litekin:
in this case ComparisonTables looks to be clearly under ranked since it has in depth comparisons for many different devices. but a single reorder to a random section doesn't really need a "discussion". when links are added, where they'll be ranked is rarely mentioned, it's just whatever looks right to nbats, which is good. if anyone thinks something should be ranked differently they can just say so, like this

flowerey:
okay then

PR #5969, opened 2026-08-04
Add CopyFonty to Unicode Text Generators

shahbakhtk8-lang:
Adding CopyFonty (https://copyfonty.com/) to the Unicode Text Generators list.

It's a free font generator that converts plain text into 500+ Unicode styles (bold, cursive, gothic, aesthetic, etc.) for use on Instagram, TikTok, Discord, and other platforms. No signup required.


nbats:

PR #5976, opened 2026-08-05
Remove Anime-Base entry from non-english.md

BennySama94:
As the Website Developer of Anime-Base.net, I can completely confirm that as of today, Anime-Base.net has closed and won't return.


nbats:
gotcha should be removed now

thank you for running it while you did

PR #5983, opened 2026-08-07
Update video.md

allflixxx:
Allflix is a sleek and modern streaming platform offering a great selection of movies, TV shows, and anime, with a clean interface designed for a smooth and enjoyable viewing experience.


nbats:
seems like a frontend for a few common sources, doesn't stand out enough sorry

PR #6009, opened 2026-08-10
Removed Portable from android and added on desktop

Lucas559-noob:
Saw it yesterday, for Android section there was a 'Portable' in description, but from what I know android don't have portable, only apk
Added portable to Desktop client, not sure if is only windows


nbats:
Thank you I will get these changes merged into the next edit

PR #6014, opened 2026-08-10
Revised list details and added thinking levels when possible

767mmy5p:
1. I'm not that confident about the formatting.
2. I couldn't figure out what the current model copilot was using.
3. Kiri told me Kimi Instant was likely Kimi K2.6.
4. Sarvam doesn't have Sarvam-30b anymore.
5. Apertus offers multiple model families once logged in so that's why I moved Apertus to another section.
6. Some of the LLMs don't work in https://aiassistantbot.pages.dev/
7. I moved Cerebras higher because the models are better than apertus and the abandoned https://aiassistantbot.pages.dev/ site plus the site's inference is way faster.
8. I moved the better reasoning note from copilot to lumo instead.

There's more stuff that's changed other than the things I've mentioned above.

### DIFF
--- a/docs/ai.md
+++ b/docs/ai.md
@@ -12,43 +12,43 @@
 
 ## ▷ Official Model Sites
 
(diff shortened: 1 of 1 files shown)


nbats:
thank you looks good to me, will merge into next update

PR #6015, opened 2026-08-10
Update mode name from 'Think Deeper' to 'Thinking Mode'

767mmy5p:
I'm moving the note to Lumo instead of copilot so that's why this is being changed.


nbats:
thank you, will merge this into the editor now

PR #6024, opened 2026-08-11
Update Claude entry with new version details

solar-bypasser:


nbats:
ty will copy into next update

PR #6043, opened 2026-08-14
Update Z.ai version in documentation

solar-bypasser:
Updated Z.ai version from GLM 5.2 to GLM 5.3.


solar-bypasser:
Only add this change when it goes online for free users.

nbats:
much appreciated, will copy into the next update

PR #6049, opened 2026-08-15
Add SeatingChartGenerator

kugua233:
Add SeatingChartGenerator under Calendars / Events, next to the existing seating-chart entry.

It is a free, no-sign-up browser-based tool for making printable wedding, party, classroom, and event seating charts.


nbats:

PR #6058, opened 2026-08-16
Update Meta AI entry with new version details

solar-bypasser:
This may be changed if I did something wrong.


nbats:
ty will copy into next update

PR #6077, opened 2026-08-19
Removed findareddit.com as it is now a gambling site, kept the subreddit link

moosetwin0:


nbats:
thank you I will merge this change into the editor

PR #6081, opened 2026-08-19
feat(educational): add country profile references (Factbook alternatives)

Sets:
**Adds one line under World Data & Statistics.**

The CIA World Factbook was shut down in February 2026, so the country-profile
slot that a lot of people used is currently empty in the wiki. Two free
alternatives:

- **[Databook](https://databook.dataint.net/en/)** — ~250 countries, 242
  indicators across 15 sections (geography, population, economy, health,
  education, environment, governance, infrastructure). Every figure names its
  source and indicator code on the page; 41 of the indicators carry the
  Factbook's own fields forward. 25 languages, no signup.
- **[OpenFactBook](https://openfactbook.org/)** — community-maintained Factbook
  successor, 260+ countries, side-by-side country comparison. Not mine, added
  because it belongs on the same line.

Both are free with no account, no trial and no paywall.

**Disclosure:** I built and run Databook. It's ad-funded, which is how it stays
free — same model as Worldometer already on that line. Nothing to buy and
nothing to sign up for. If you'd rather list only OpenFactBook, that's a fair
call and the line still works.

### DIFF
--- a/docs/educational.md
+++ b/docs/educational.md
@@ -698,6 +698,7 @@
 * [Data Commons](https://datacommons.org/) or [Engaging Data](https://engaging-data.com/) - World Data Visualizations
 * [EverySecond](https://everysecond.io/) - Visualize Per Second World Data
 * [If It Were My Home](https://www.ifitweremyhome.com/), [GlobalEdge](https://globaledge.msu.edu/) or [MyLifeElsewhere](https://www.mylifeelsewhere.com/) - Country Data Comparisons
+* [Databook](https://databook.dataint.net/en/) or [OpenFactBook](https://openfactbook.org/) - Country Profiles / World Factbook Alternatives
 * [NationsEncyclopedia](https://www.nationsencyclopedia.com/) or [⁠CityPopulation](https://citypopulation.de/) - Location / Population Data
 * [City Data](https://www.city-data.com/) - US City Data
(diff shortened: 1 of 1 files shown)


Sets:
Added a second line while I was here — a companion reference for trade and logistics:

**[Logibook](https://logibook.dataint.net/en)** — 58,730 entities across 250 countries: the Harmonized System (5,613 subheadings with CN/TARIC/US-HTS crosswalks), dangerous goods (2,347 UN numbers with ADR/RID/ADN/IMDG/ICAO Table A side by side), 3,316 ports, 6,385 airports with runway and frequency detail, 5,290 trade zones, Incoterms 2020, and code registers (ISO 6346, IATA ULD, UNECE Rec 20/21). Free, no signup, 25 languages, every figure attributed to its source dataset.

Same disclosure as above — I build and run it, ad-funded. Happy to split this into a separate PR if you'd rather keep them apart, or drop it if trade/logistics is out of scope for the educational page.

nbats:
Thank you for sharing with us Sets I appreciate it, I moved all 3 sites including all of the info you provided to our Discord for testing

Sets:
Thanks for taking the time to look at it, and for moving it through your process rather than just closing it.

One thing that may be useful for testing: Databook and Logibook are the same project's two halves, so if one fits the wiki and the other doesn't, splitting them is completely reasonable. Logibook is the narrower of the two — it's aimed at trade/logistics work rather than general reference.

No follow-up needed from me; I won't chase this. If the testers hit anything wrong in the data, I'd genuinely like to hear it.

PR #6100, opened 2026-08-22
Add ReelSuites to video streaming list

Solonce:


nbats:
seems to be a single common source, it wont make it in as is

it'll need more servers

or a custom player that checks sources

or a unique source

Solonce:
Made a new pull request with changes @nbats

PR #6123, opened 2026-08-24
Typos, Fixes, and Cleanup

bbbreaddd:
Mostly just Grammar fixes, punctuation, small rewording,  removing extra spaces, fixing broken links, and deleting duplicate links

I also found out that MagicNZB likely doesn't have a free trial anymore. I made an account and unless I missed something, there wasn't any free option.
I also added some changes to the lint-markdown script to better catch these issues and reduce false positives


flowerey (approved):
lgtm 👍

PR #6264, opened 2026-09-08
Fix links v2

zinklog2:
might have some issues since i didn't create a branch before, but from future these PRs should be pretty seamless


FMHYedit
saberofxebecc, line 1: probably unintentional, but I think your branch contains a nested submodule, which caused this to point to it

PR #6334, opened 2026-09-13
Update text-tools.md

abdulsuboor780:


abdulsuboor780 (commented):
Added CaratteriSpeciali to the Text Tools section as a free resource for generating and copying stylish fonts, symbols, and special characters. The site is free to use and requires no registration.

PR #6360, opened 2026-09-15
Refine categories for download resources

Axekinn:
Updated resource categories for various download sites to include more specific content types.


nbats:
sorry but a lot of these don't match the way we tag things, we also generally don't tag software on these general DDL sites or games as safety is questionable

PR #6374, opened 2026-09-17
Update DnD-Books link and description

nazdridoy:
DnD-Books includes premade character sheet, and DnD 5.5e core books


nbats:
looks good to me thank you for keeping it updated

I'll adjust this all in the editor now

PR #6377, opened 2026-09-18
fix: correct spelling of 'Docker Hub' in developer tools section

AryanBhanot:
The entry was named “Hub Docker” while the official name is “Docker Hub"


nbats:
thank you I'll get it fixed in the editor now

Issue #6380, opened 2026-09-18
Could you please update our listing to use .to and .st as our primary domain and .link for mirrors?


hackax-1:
Hi @nbats, thank you for adding us to the wiki!

Our previous .watch and .ws domains are no longer active. Could you please update our listing to use .to and .st as our primary domains and .link for mirrors?

Current listing:

* [Reedstreams](https://reedstreams.watch/), [2](https://reedstreams.ws/) - Stream Aggregator / [Status](https://reedstreams.link/)

Updated listing:

* [Reedstreams](https://reedstreams.to/), [2](https://reedstreams.st/) - Stream Aggregator / [Mirrors](https://reedstreams.link/)

nbats:
yes thank you for letting us know, I will adjust it in the next edit

PR #6381, opened 2026-09-18
update Reedstreams domains

nqlp:
`.watch` and `.ws` are dead, so this swaps Reedstreams over to `.to` and `.st` — that's what the site folks asked for in #6380.

Also folded the old `Reedsports` entry into it, since it was just pointing at `reedstreams.to`, which is now the main Reedstreams link. No point having the same domain listed twice.

Left the `.link` one as `[Status]` for now. The issue asks for it to say `[Mirrors]` instead — happy to change it, just say the word.

Ref #6380


nbats:
thank you it was also sent in another thread, will update soon

PR #6387, opened 2026-09-19
Add GrabGPU

taiberium:
```
* [GrabGPU](https://grabgpu.com/) - Compare Cloud GPU Prices / Stock
```

Added next to GetDeploying. GrabGPU compares hourly GPU rental prices across 22 cloud providers and 71 GPU models, and shows the stock each provider reports, so you can see what is actually rentable. No account, no ads, no paywall. Data comes from four public provider APIs and is refreshed daily.


nbats:

